PR #2, opened 2019-05-15
Add third party libraries

KarthikRIyer:
This PR adds the AGG and lodepng libraries. I haven't called one of them in the executable at the moment because both are c++ libraries.


BradLarson:
Quick question about these: the overall project is licensed as Apache 2.0. Are there any conflicts with the AGG or lodepng licenses and the overall license triggered by including them directly in the repository?

Would there be any advantage to having AGG+wrapper and lodepng+wrapper as separate repositories that are then pulled in as dependencies? Is there a possible mechanism for building the project such that these are not included on incompatible platforms?

KarthikRIyer:
I don't think there is a problem with AGG. The website for AGG has expired But someone had put up the code for AGG on github. I found this there:

The Anti-Grain Geometry Project
A high quality rendering engine for C++
http://antigrain.com

Anti-Grain Geometry - Version 2.4
Copyright (C) 2002-2005 Maxim Shemanarev (McSeem)

Permission to copy, use, modify, sell and distribute this software
is granted provided this copyright notice appears in all copies.
This software is provided "as is" without express or implied
warranty, and with no claim as to its suitability for any purpose.

BradLarson:
Found this in [Wikipedia](https://en.wikipedia.org/wiki/Anti-Grain_Geometry):

> While AGG version 2.5 is licensed under the GNU General Public License, version 2 or greater, AGG version 2.4 is still available under the 3-clause BSD license and is virtually the same as version 2.5.

Sounds like 2.5 would definitely be out, but 2.4 is a better story. For compatibility of 3-clause BSD vs. Apache 2.0, I see this: [Is BSD license compatible with Apache?](https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache). Sounds compatible, but still wondering if a separate project for just the AGG wrapper would make sure that everything was clean.

I still like the idea of potentially enabling / disabling parts of the rendering backend on various platforms. For example, you might want to selectively disable AGG as a backend for iOS in favor of the native Core Graphics so that the bulk of the AGG library wasn't added to an iOS application where it wasn't needed. Don't know if that would be better enabled via an external project for AGG or just selective enabling or disabling of internal dependencies.

KarthikRIyer:
lodepng is under zlib license. Apache 2.0 allows patent use whereas zlib doesn't. Maybe we could add a zlib license for lodepng?

KarthikRIyer:
What I'm doing currently is writing a class using AGG to draw lines, rectangles, etc and then wrapping it. I don't think it makes sense to have a separate project for AGG. We also need the wrapper to follow the Renderer protocol, how would that work if we have a separate project?

BradLarson:
This seems like a good reference for what's compatible with Apache 2.0: https://apache.org/legal/resolved.html#category-a

It explicitly lists zlib and BSD 3-clause as compatible, with comments about adding attribution to the appropriate NOTICE file.

Still, we can keep thinking about whether it makes sense to have these as separate things pulled in via Swift Package Manager (or external Xcode projects), at least from an organizational perspective. If someone wants to add a new rendering backend, would that be done via an addition to this core project or as a separate project that could be selectively pulled in? Don't know how others have approached this.

KarthikRIyer:
matplotlib includes agg in the main project. I can also see two other libraries, ttconv and libqhull in the extern folder of the main project. I can't see any other renderers included.

PR #3, opened 2019-05-15
Add renderer backends

KarthikRIyer:
I have added the renderer backends as suggested with them following a Renderer protocol and eliminating Vectorizer and VectorizerProtocol. 
I've removed the save image methods from the protocol. 

Please review. I'll open the next two PRs to add the LinePlot and examples once the first two are merged.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,22 @@ let package = Package(
             dependencies: [],
        	    path: "framework/lodepng"),
 	.target(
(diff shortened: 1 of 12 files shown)


BradLarson (commented):
I commented about this on the other PR, but I think the Renderer instances should probably be classes instead of structs due to the amount of internal mutation involved with them. That would remove a lot of the mutating references here and generally clean up the code.

PR #4, opened 2019-05-16
Add LinePlot

KarthikRIyer:
I've added LinePlot.
I've also separated what I thought to be common plot properties like PlotBorder, PlotDimensions, PlotLabel, etc. into separate types and put them under **Util**. 
I've also separated the plotting logic in an extension.


framework/Util/Point.swift
```diff
@@ -1,9 +1,10 @@
 // class defining a point
 public struct Point {
-  public var x : Float
-  public var y : Float
+  public var x : Float = 0
+  public var y : Float = 0
```
BradLarson:
```suggestion
  public let x: Float
  public let y: Float
```

I'd make these let, because you're generally not going to want to mutate them inside of the Point. Instead, you'd perform arithmetic on the Point itself to generate a new Point.

framework/Util/Point.swift
```diff
@@ -1,9 +1,10 @@
...
   public init(_ x : Float, _ y : Float){
     self.x = x
     self.y = y
   }
+  public init() {}
```
BradLarson: I'd remove the blank initializer for zero, in favor of something like this:

```
public static let zero = Point(0.0, 0.0)
```

That way, you can call Point.zero for a 0, 0 point and that becomes clear from the code itself.

BradLarson (approved):
I like how this is shaping up, with the higher-level organization logic for the plot being descriptive and relatively simple in Swift, split from the rendering code via an extension. I think that nicely splits the concerns and makes the overall code easier to read and extend.

I made a couple of inline comments about the simpler types, and you can look here for an example of what I'm talking about:

https://github.com/BradLarson/GPUImage2/blob/master/framework/Source/Position.swift

For simple types like Point, you might want to define some operator overloads for addition, subtraction, division or multiplication by a scalar, etc., but those can come later. The `public static let` pattern is common for regularly-used values, and here's a list of those I used for Color that you might want to copy over to your respective type for that:

https://github.com/BradLarson/GPUImage2/blob/master/framework/Source/Color.swift

I know Marc had suggested the Renderers possibly be structs, but I'm thinking that given the amount of internal mutation within them they should be classes instead. While using a reference type could potentially open up unintended coupling between remote parts of code using the same Renderer instance, I think it makes sense here. That would allow you to remove the inout, &, mutating markers on the protocol methods and other interactions with the renderer, really cleaning up that code.

KarthikRIyer:
I've made the changes. Are the PRs other than the examples ready to merge?

BradLarson:
This all looks good to me, I approve pulling it in. Any changes / additions to this can easily come afterward.

PR #5, opened 2019-05-16
Add LineChart examples

KarthikRIyer:


BradLarson (commented):
Generally, when I've seen example applications structured with the Swift Package Manager in mind, you usually have a directory named after the application and within it a main.swift file. In this case, maybe a separate SingleSeriesLineChart and MultipleSeriesLineChart directory with main.swift files in both.

The nice thing with that is that main.swift files are special and they execute in order from the top of the file down, so you could remove your func xxx {} wrapping and just have the plotting commands directly inline for both examples.

KarthikRIyer:
I did this because I wanted to keep all the line chart examples together. Is there a disadvantage in the way it's currently done? @BradLarson

BradLarson:
There isn't a huge difference in terms of code, just a matter of organization. I personally like to make small test applications that have a single purpose, but there aren't any hard rules about that.

Underscored functions are rare in Swift, so stylistically I might rename these to a camelcase variant that states what they do, but again we can tweak these examples after they're included in the repository.

Issue #8, opened 2019-05-19
Add basic SubPlot support


KarthikRIyer:
Add basic SubPlot support with the graphs stacked horizontally and vertically

PR #9, opened 2019-05-19
Add SubPlots for SVG, add opacity to SVG Renderer, update examples

KarthikRIyer:
#8 
I've added  SubPlot support where one can add multiple SubPlots and stack them vertically, horizontally or in a grid.
If one needs to plot just a single graph, the old method will still work.
Currently SubPlots are supported only for SVG, because I'm having some trouble changing the buffer size according to the plot dimensions.

In addition to this I've added opacity to SVG which wasn't there previously.

I've also made a project for milestone 1 to keep track of the progress.


framework/LinePlot/LinePlot.swift
```diff
@@ -115,7 +143,7 @@ extension LineGraph{
   }
 
   // functions implementing plotting logic
-  func calcLabelLocations(renderer renderer : Renderer){
+  func calcLabelLocations(renderer renderer : inout Renderer){
```
BradLarson, line 146: Is there a reason for the switch back to `inout` on the Renderer interfaces? Is this pulling from older code where that had been the case? I think it still makes sense to treat the Renderers as reference types, and the interface was nicely cleaned up through the removal of the `inout`, etc. in one of the more recent commits.

KarthikRIyer: I wanted to pass the renderer as reference to the plot so that, when I am plot ring subplots, I can pass the same renderer object to all the plot objects and in the end just save the output.

KarthikRIyer: Isn't passing by reference better in this case?

BradLarson: If your Renderer is a reference type (a class), it should be natively passed by reference, even through to subplots. This may require slightly changing the definition of Renderer to 

```
public protocol Renderer: AnyObject {
```

so that it's understood that Renderer will always be a reference type. Thinking about it more, having Renderer as a value type could lead to problems with state changes among sub-operations, so I think we'll always want it to be a reference type.

KarthikRIyer: Ok. I'll make the changes.

KarthikRIyer: Other than the inout issue is this ready to merge? I'll think I can take a look at the AGG buffer issue later, independently, because the SubPlot logic is independent, therefore it'll stay the same.

I have removed inout, and am including the changes in the PR for functions, which I should be able to open by tonight.

BradLarson: If you'll make the value->reference type changes later, or in a separate pull request, I'm fine with pulling this in.

Issue #10, opened 2019-05-20
Add support to plot functions in LineGraph


KarthikRIyer:

BradLarson:
As a comment on the design of this, it'd be great to have this use a function / closure as a parameter for this. Something of the type `(Numeric) -> (Numeric)` for an f(x), for example. This would allow for arbitrarily complex functions to be passed in. Might have to think about the signature for that, to account for parametric functions with more than one Y value at a given X, though.

You'd have to account for what happens in NaN cases, and may need to think about what numeric types to support for this.

You'd also maybe think about whether to have these in parallel to the Series array that's currently present in the LinePlot (as a separate array of Function2D or whatever), or if you'd want to create an encompassing type or protocol that could include both Series and functions. Something that would provide values in response to queries for a specific X coordinate, maybe. That's all something that could be architected when you come to it, however.

KarthikRIyer:
I have overloaded the addSeries  function with (Float)->Float as a parameter at the moment. I'll get simple functions to plot first and then move on to multiple Y and NaN cases. 
I am also accepting the number of samples also as an Int.
Does that seem fine?

But first I'll have to change the plotting logic. I don't know how or why, but I forgot to consider negative values😅  So now the line goes below the x axis and out of the image when you give negative values.

BradLarson:
That sounds like a good plan for now. Best not to over-architect things initially, and build or revise interfaces as they are needed for different capabilities.

Yeah, the lack of negative support is a great example of why the little test applications are so useful to have early on. I always forget something, and building against the test applications and checking their output helps to point out errors. It's not really test driven development, but it's in the same ballpark.

PR #11, opened 2019-05-21
Add basic function plotting feature

KarthikRIyer:
#10 
I've added functionality to plot simple functions.
I've ignored values that are NaN or inf.

Also in cases like **plotting 1/x**, depending on the range of x, the **first positive y value after x=0.0 was something like 60000**, and the **first negative value before x=0.0 was something like -20**. On plotting this, as the positive y is much larger than the negative y, the plot seemed like a single peak at zero and no negative peak for negative x.

I wasn't sure how to handle this so I capped the **y** value to a max and min of **10 times the range of x**. 

Should we consider converting all the Floats to a generic, like the one in the Cap function under Util?


framework/Util/Cap.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+public func cap<T :Numeric & Comparable>(value v : T, min a : T, max b : T) -> T {
```
BradLarson: I've commonly seen this function named `clamp` in GLSL and other libraries. Here's a possibly simpler definition of that:

https://stackoverflow.com/a/46799935/19679

```
public func clamp<T>(_ value: T, minValue: T, maxValue: T) -> T where T : Comparable {
    return min(max(value, minValue), maxValue)
}
```

KarthikRIyer: Oh yes, I remembered the name wrong. I'll rename it. But is this the best way we can deal with such a situation?

framework/LinePlot/LinePlot.swift
```diff
@@ -62,50 +62,71 @@ public class LineGraph : Plot {
     let s = Series(points: pts, label: l, color: c)
     series.append(s)
   }
+  public func addSeries(function f : (Float)->Float, x1 a : Float, x2 b : Float, numberOfSamples numberOfSamples : Int = 400, label l: String, color c : Color = Color.lightBlue) {
```
BradLarson: Generally in Swift, parameter names tend to align with their internal names unless the signature can be defined in a better way by having them be different. Also, I tend to like having descriptive variable names. Single-character variable names are something I tend to avoid.

When it comes to whitespace around colons, the [Google Style Guide](https://google.github.io/swift/) and standard Swift style is to have no whitespace before the colon, and one space afterward for parameters. In my projects, you might notice that I eliminate whitespace before and after colons, but that's my own idiosyncratic style that I'm gradually working myself out of. I lost that argument with others in the Swift community early on.

I might suggest a function definition like the following:
```
public func addSeries(function: (Float)->Float, xMin: Float, xMax: Float, numberOfSamples: Int = 400, label: String, color: Color = Color.lightBlue) {
```

with the appropriate modification in the following function to support those internal parameter names. This is something that could be applied across all functions, and I think it would make them both easier to read and easier to understand the internal logic.

BradLarson: Also, would addFunction() be a better name than addSeries(), in order to set this apart? In that case, you maybe could change addSeries to:

```
public func addSeries(_ series: Series){
```

and addFunction to:

```
public func addFunction(_ function: (Float)->Float, xMin: Float, xMax: Float, numberOfSamples: Int = 400, label: String, color: Color = Color.lightBlue) {
```

making the first arguments implicit in the title of the function.

framework/Util/Cap.swift
BradLarson: This is a way I've done clamping in the past. It simply requires a type to be Comparable, so that's pretty broad, and the use of generics prevents type erasure when using the function. Should be pretty good for all our needs.

It's a simple enough function that I might put it in a general HelperFunction.swift or something like that. I imagine there will be a few other functions that might be useful to throw in there.

KarthikRIyer: I didn't mean the way to do the clamping.
What I meant was to prevent extremely large values from ruining the plot.

framework/LinePlot/LinePlot.swift
KarthikRIyer: Now that I think about it, addFunction does seem better.

KarthikRIyer: I'm sorry about this, I know you had mentioned it earlier. I have this bad habit of writing non-descriptive variable names and later get confused when I revisit the code. I'll take extra and deliberate care from now on.

framework/Util/Cap.swift
BradLarson: That's a good question. clamp's an easy way to handle it at first, given a display range, but you might want to show the discontinuity in the plotted function at the top or bottom of the graph as they extend upward.

Might be worth looking into how other plotting frameworks handle this, because I can't remember how we dealt with things like asymptotes before.

KarthikRIyer: Alright, I'll search for it. I plotted 1/x in matplotlib, I didn't see the code, but from the graph it seemed like they were skipping the infinity points and then clamping the large values. I thought this because of a near vertical line at x = 0

BradLarson: If the values were clamped to ranges that were just offscreen, you might be able to show the appropriate discontinuities on the edges and potentially draw the lines that connect really large values to really small ones. That could be a simple solution that looks nice onscreen.

KarthikRIyer: The problem here would be what value would I clamp it to? I'm just querying for y values for each x. Then I decide the x and y ranges according to maxX maxY minX minY. When I don't clamp and the absolute maxX and minY have a large difference, we won't be able to se either one of them. This explanation seems really sloppy. I think I'll be able to explain the problem better with an image. I'll do that when I get home.

KarthikRIyer: **with clamping**
_06_function_plot_line_chart
**without clamping**
_06_function_plot_line_chart

This is 1/x

r -5.000002
r -5.714288
r -6.66667
r -8.000005
r -10.000008
r -13.333346
r -20.000029
**r -40.00012**
**r 13421773.0**
r 39.99988
r 19.99997
r 13.33332
r 9.999992
r 7.999995
r 6.666663
r 5.7142835

These are the values of the function close to 0
max Y = 13421773
min Y = -40
So in comparison to maxY minY is very small.

KarthikRIyer: @BradLarson It seems matplotlib has the same problem like above
I tried plotting for different **x** ranges and saw the same problem.
But when I tried for just positive or negative numbers they were clamping at the inverse of the step value.
Like:
```
import matplotlib
import matplotlib.pyplot as plt
import numpy as np

t = np.arange(-2.0, 0.0, 0.001) #here the graph is clamped at 1000 or -1000 whichever is applicable
s = 1/t

fig, ax = plt.subplots()
ax.plot(t, s)

ax.set(xlabel='time (s)', ylabel='voltage (mV)',
       title='About as simple as it gets, folks')
ax.grid()

plt.show()
```

I think we too can clamp at the inverse of the step value.

BradLarson: That makes sense. If the user manually fixes a range for the viewport of the graph, you could easily clamp to that, but following matplotlib's behavior is sensible in the case you describe.

Thanks for doing the legwork to research this, I couldn't remember how we handled this case before in Core Plot.

BradLarson (approved):

PR #15, opened 2019-05-28
Rework package structure to include all plots and util code in …

KarthikRIyer:
…SwiftPlot, and have separate modules for each renderer


examples/LineChartMultipleSeries/main.swift
```diff
@@ -1,6 +1,7 @@
-import LinePlot
-import Util
-import Renderers
+import SwiftPlot
+import SwiftPlot
```
BradLarson: Looks like this accidentally was added twice.

BradLarson (approved):
Aside from the one accidental duplication noted above, I like this structure. You could even think about hiding the CAGGRenderer and CSVGRenderer helper C modules inside of the AGGRenderer and SVGRenderer subdirectories (maintaining the appropriate dependencies there) to contain everything needed for each specific renderer in one place.

KarthikRIyer:
Do you mean something like this:

dir AGGRenderer
- dir CAGGrenderer
        -dir include
                -CAGGRenderer.h
        -CAGGRenderer.cpp
- dir AGGRenderer
        -AGGRenderer.swift

And in Package.swift: 

```swift
.target(
            name: "CAGGRenderer",
            dependencies: ["CPPAGGRenderer"],
    	    path: "framework/AGGRenderer/CAGGRenderer"),
  .target(
            name: "AGGRenderer",
            dependencies: ["CAGGRenderer","SwiftPlot"],
            path: "framework/AGGRenderer/AGGRenderer"),
```

BradLarson:
Yup, I think that's close to what I was thinking about. If you wanted to, you could move the AGGRenderer.swift file up a level to the main AGGRenderer subdirectory and then use `, exclude:["framework/AGGRenderer/CAGGRenderer", "framework/AGGRenderer/CPPAGGRenderer"]` for the AGGRenderer target to manually exclude the C and C++ modules. That'd save a subdirectory.

Don't have to, though. That's just a potential minor simplification.

Issue #12, opened 2019-05-22
Add multiple independent y axes feature to LineGraph


KarthikRIyer:

KarthikRIyer:
For this I was thinking we can have a primary and secondary axis in LinePlot. While adding the series we can assign which axis we wan to use, which would be the primary one by default. For differentiating which axis a series belongs to we can maybe make the lines dashed...
Any ideas?

In matplotlib, the new axis is twinned with an existing axis, so the new axis is plotted opposite to the existing axis. But here each axis has a series of data. 
Something like this:
```python
import numpy as np
import matplotlib.pyplot as plt

fig, ax1 = plt.subplots()
t = np.arange(0.01, 10.0, 0.01)
s1 = np.exp(t)
ax1.plot(t, s1, 'b-')
ax1.set_xlabel('time (s)')
# Make the y-axis label, ticks and tick labels match the line color.
ax1.set_ylabel('exp', color='b')
ax1.tick_params('y', colors='b')

ax2 = ax1.twinx()
s2 = np.sin(2 * np.pi * t)
ax2.plot(t, s2, 'r.')
ax2.set_ylabel('sin', color='r')
ax2.tick_params('y', colors='r')

fig.tight_layout()
plt.show()
```
In ggplot, from what I could make out, the secondary axis is plotted opposite to the primary axis but it's scale has to be set manually, by multiplying or dividing the primary scale values by another value.
Something like this:
```R
p <- ggplot(obs, aes(x = Timestamp))
  p <- p + geom_line(aes(y = air_temp, colour = "Temperature"))
  
  # adding the relative humidity data, transformed to match roughly the range of the temperature
  p <- p + geom_line(aes(y = rel_hum/5, colour = "Humidity"))
  
  # now adding the secondary axis, following the example in the help file ?scale_y_continuous
  # and, very important, reverting the above transformation
  p <- p + scale_y_continuous(sec.axis = sec_axis(~.*5, name = "Relative humidity [%]"))
  
  # modifying colours and theme options
  p <- p + scale_colour_manual(values = c("blue", "red"))
  p <- p + labs(y = "Air temperature [°C]",
                x = "Date and time",
                colour = "Parameter")
  p <- p + theme(legend.position = c(0.8, 0.9))
p
```

KarthikRIyer:
Any ideas?

PR #16, opened 2019-05-29
Add secondary axis feature

KarthikRIyer:
#12 
I've added a secondary axis. The user can specify the axis while adding the series like so:
```swift
lineGraph.addSeries(x, y, label: "Plot 2", color: Color.orange, axisType: Axis.SECONDARY_AXIS)
```
By default the axis is the primary one.
To differentiate between the series for both axes, I've made the series for secondary axis dashed.

I've also added an overload for the `addSeries` function with just the y-values passed in. In this case the x-values will be [1, 2, 3, 4, ...]

Once this is merged, I think the main stuff for the 1st milestone will have been covered as discussed in the email:

> To summarise what I can do for milestone 1:
> - Make a simple Line Chart imaplementation with SubPlot support.
> - Setup tests that saves images next to the expected results.
> - Get a base64 encoded PNG and use it in Jupyter notebook. Later work on python free implementation.
> - Complete line chart implementation in the leftover time.

@marcrasi  For the python free implementation to display plots could you shed some light as to how swift-jupyter works, and how I can approach the problem? Or some resources I can read from regarding this?

Then I was thinking of writing some documentation for the work done till now. 
I think I can keep the Bar Chart work in the 1st milestone itself. I'll start working on it simultaneously finalising this PR.


marcrasi:
Does plotting work in jupyter now? If so, you should make a post on https://forums.fast.ai/c/harebrain soon!

I'll respond to your question about drawing without python in a github issue.

KarthikRIyer:
@marcrasi I haven't tested it in jupyter yet. I'll do that first thing today. If I should post it, on the fast.ai forum, I think it would be better if I had some basic documentation ready. So I think I should write some basic documentation and then post it. Does that seem ok?

marcrasi:
Yep!

KarthikRIyer:
Just out of curiosity, how does swift work in colab?  Can I get plotting to work there?

BradLarson:
@KarthikRIyer - Swift should work in Colab as it does in plain Jupyter, with the occasional slight lag in what Swift for TensorFlow version is used in Colab vs. a local toolchain. I think you still need to start from an existing Swift notebook as a template, since you can't create one from scratch yet, but after that it should behave like a local Jupyter notebook.

You may need to put an arbitrary tag on the project (0.0.1 or whatever) to have an initial version to refer to when pulling it into a Colab notebook. Here's a sample Fast.ai notebook that imports some Swift modules (first cell there):

https://github.com/fastai/fastai_docs/blob/master/dev_swift/00_load_data.ipynb

If you can, I'd recommend creating an examples/Notebooks directory to host example Jupyter notebooks that ship with the project, and you ignore that directory in the Package.swift (these don't need to be compiled). You should then be able to import the local project based on relative paths, instead of calling out to GitHub in those examples. Those would provide both test cases for the Jupyter support and examples for how to use it.

KarthikRIyer:
@BradLarson I tried pulling SwiftPlot into colab and using it...but the display function that I added to swift-jupyter isn't available on colab.

I've added a sample notebook in a direct commit to master. I'll move it to a folder named Notebooks.

I've already added a tag 0.0.1

KarthikRIyer:
@BradLarson @marcrasi could you please review this? It would be nice to post on fast.ai forum with a complete Line Graph implementation. Only this part is left.

framework/SwiftPlot/Axis.swift
```diff
@@ -0,0 +1,10 @@
+public class Axis{
+  public var scaleX: Float = 1
+	public var scaleY: Float = 1
+  public var plotMarkers: PlotMarkers = PlotMarkers()
+  public var series = [Series]()
+  public init(){}
+
+  public static let PRIMARY_AXIS: Int = 0
+  public static let SECONDARY_AXIS: Int = 1
+}
```
BradLarson, line 12: In Swift, even static constants tend not to be all-uppercase like in C-family languages. I'd name these like you would a property (lowercase first letter, camelcase after that).

BradLarson: As a broader design issue, I'd make the primary / secondary axis a true enum and not a hardcoded numeric value. You want to make these cases exhaustive, and starting to work with numbers for them opens you up to the issues inherent with numerical enums elsewhere.

This could be an enum type contained within Axis, such as AxisLocation (or just Location, because its namespace would be Axis.Location as a subtype) or outside of it if other types would use it. I'd do it inside Axis for now, and see how that works.

KarthikRIyer:
@BradLarson PR updated

examples/LineChartSecondaryAxis/main.swift
```diff
@@ -16,8 +16,8 @@ var svg_renderer: SVGRenderer = SVGRenderer()
 var plotTitle: PlotTitle = PlotTitle()
 
 var lineGraph: LineGraph = LineGraph()
-lineGraph.addSeries(x1, y1, label: "Plot 1", color: Color.lightBlue, axisType: Axis.PRIMARY_AXIS)
-lineGraph.addSeries(x, y, label: "Plot 2", color: Color.orange, axisType: Axis.SECONDARY_AXIS)
+lineGraph.addSeries(x1, y1, label: "Plot 1", color: Color.lightBlue, axisType: Axis.Location.primaryAxis)
+lineGraph.addSeries(x, y, label: "Plot 2", color: Color.orange, axisType: Axis.Location.secondaryAxis)
 plotTitle.title = "SECONDARY AXIS"
```
BradLarson, line 21: As an additional simplification, because Swift can infer types, I often leave off the type prefix for enum values. `.primaryAxis` instead of `Axis.Location.primaryAxis` or `.lightBlue` instead of `Color.lightBlue`. If type can't be inferred, or if specifying the type prefix makes things clearer, then I'll do it but otherwise I tend to drop that to focus on the value.

KarthikRIyer: Done. I also made the stackPattern in subplots an enum.

BradLarson (approved):
This looks good to me now.

PR #21, opened 2019-06-01
Implement basic bar chart feature

KarthikRIyer:
#17 
I've added a simple bar chart where the user adds a series using the Point that was previously used in LineGraph. Here the x can be a Float or String. Both will work.

Currently I'm thinking of a way to implement hatching of the bars.


KarthikRIyer:
I've added hatching. There are two patterns at the moment. Lines at +45 and -45 degrees. I'll see what other patterns I can implement.

framework/SwiftPlot/Renderer.swift
```diff
@@ -5,7 +5,8 @@ public protocol Renderer: AnyObject{
 	var plotDimensions: PlotDimensions { get set }
 
 	func drawRect(topLeftPoint p1: Point, topRightPoint p2: Point, bottomRightPoint p3: Point, bottomLeftPoint p4: Point, strokeWidth thickness: Float, strokeColor: Color)
-	func drawSolidRect(topLeftPoint p1: Point, topRightPoint p2: Point, bottomRightPoint p3: Point, bottomLeftPoint p4: Point, fillColor: Color)
+	func drawSolidRect(topLeftPoint p1: Point, topRightPoint p2: Point, bottomRightPoint p3: Point, bottomLeftPoint p4: Point, fillColor: Color, hatchPattern: Series.hatching)
+	func drawSolidRectTransformed(topLeftPoint p1: Point, topRightPoint p2: Point, bottomRightPoint p3: Point, bottomLeftPoint p4: Point, fillColor: Color, hatchPattern: Series.hatching)
```
marcrasi: What's the difference between `drawSolidRect` and `drawSolidRectTransformed`? Some comments on the methods in this protocol would be helpful.

Is there some way that you could do everything you need with only one function, so that renderer implementations don't have to implement as much logic?

framework/SwiftPlot/Series.swift
```diff
@@ -1,13 +1,21 @@
 // struct defining a data series
 public struct Series {
 
+	public enum hatching: Int, CaseIterable{
+		case none = 0
+		case forwardSlash = 1
+		case backwardSlash = 2
+	}
+	public var hatchPattern: hatching = .none
```
marcrasi: could things that are plot-specific (like hatchPattern is specific to bar chart) go into some plot-specific "series options" datastructure?

e.g. `BarChartSeriesOptions { public var hatchPattern: hatching }`

framework/SwiftPlot/Series.swift
```diff
@@ -1,13 +1,21 @@
 // struct defining a data series
 public struct Series {
 
+	public enum hatching: Int, CaseIterable{
```
marcrasi: type names should always be capitalized in swift style

framework/SwiftPlot/SubPlot.swift
```diff
@@ -1,7 +1,10 @@
 public class SubPlot{
-	public static let VERTICALLY_STACKED: Int = 0
-	public static let HORIZONTALLY_STACKED: Int = 1
-	public static let GRID_STACKED: Int = 2
+
+	public enum stackPattern {
```
marcrasi: capitalized

framework/SwiftPlot/BarChart.swift
```diff
@@ -0,0 +1,297 @@
+import Foundation
```
marcrasi, line 1: the indentation is showing up pretty weirdly for me in this file (and some of the others in this PR). there are tabs. I think if you change them to 4-space indent it'll fix it.

framework/AGGRenderer/CPPAGGRenderer/CPPAGGRenderer.cpp
```diff
@@ -81,24 +89,110 @@ namespace CPPAGGRenderer{
...
+        span_gen_type sg(img_src, 0,0);
+        sg.alpha(span_gen_type::value_type(255.0));
+
+        m_ras.add_path(rect_path);
+        agg::render_scanlines_aa(m_ras, m_sl_p8, rb_pre, sa, sg);
+      }
+
+    }
 
+    void generate_pattern(float r, float g, float b, float a, int hatch_pattern){
```
marcrasi, line 103: could this hatching be implemented in Swift, in terms of other renderer primitives?

an advantage of doing that would be that if you add later renderers, you can reuse the hatching code for them.

KarthikRIyer: @marcrasi I too had thought of that. Both AGG and SVG provide ways of filling up shapes with patterns which I why I went for that. When filling patterns with shapes we will need to handle the clipping of the pattern at the shape boundaries and maybe other hurdles. In case of SVG I think using primitives for hatching will surely make viewing the image slower because of increased parsing time because of large no. of statements to draw each primitive individually. 

Not that we need to do it this way but matplotlib also doesn't support hatching on all backends.

What do you think?

framework/SwiftPlot/BarChart.swift
KarthikRIyer: I'll correct this using the formatter

framework/SwiftPlot/Renderer.swift
KarthikRIyer: All the functions with the Transformed suffix use a coordinate system with the shifted origin which is shifted to the bottom left of the graph border. Maybe Transformed isn't the best suffix to indicate that. 

I can write all the code in one function with a boolean parameters decide if we want to shift the origin or not, if that seems alright?

framework/SwiftPlot/Series.swift
KarthikRIyer: I'll make the change

KarthikRIyer:
@marcrasi made the changes apart from moving Hatching to Swift

framework/AGGRenderer/CPPAGGRenderer/CPPAGGRenderer.cpp
marcrasi: Yeah, this makes sense!

marcrasi (approved):

KarthikRIyer:
@marcrasi I'll try adding these patterns too and then merge
'|', '-' , '+' ,  'x'

Then I'll implement horizontal bar chart and stacked bar chart

PR #23, opened 2019-06-06
Implement horizontal and stacked bar chart

KarthikRIyer:
#22


KarthikRIyer:
@BradLarson @marcrasi 
Can you please review this?

BradLarson (approved):
This all looks good to me.

KarthikRIyer:
Thanks for the review. Merging.

KarthikRIyer:
@BradLarson what if we require just the string variable in case of using words in bar chart? If we use computed properties for string, we'll have to check if while initializing the string or float ones were initialized. To do that we'll have to write a custom getter like this: https://stackoverflow.com/a/49162274/9126612
This doesn't seem very useful...

If we do not want both string and float to be stored simultaneously we can maybe create a separate StringPoint or BarGraphPoint type?

BradLarson:
@KarthikRIyer - This is where it might be useful to start thinking about how to provide data for graphs and plots beyond pairs of Float values. If bar charts will sometimes need Strings, sometimes Ints, and the others might want to take in Doubles, Decimals, or other numeric types (I could see taking in Tensor pairs at some point, as a Swift for TensorFlow-specific addition), maybe it's time to contemplate a different input architecture beyond Points.

That's a larger discussion, and might be worth an issue of its own. Would switching to generics be the right path for this? Maybe making plots have their input data comply to specific protocols (representable as pairs of Floats for scatter plots, representable as pairs of String-Float values for bar charts)? Do you still want to explicitly provide pairs, or do you want to switch to having an independent Series for each axis?

KarthikRIyer:
What if we make the variables inside point as generics, and accept (String, Numeric generic) in bar chart and (Numeric, Numeric) in Scatter Plot, Line Chart, etc? But in this case when we use the helper functions like getMaxX which uses > operator on x in Point, the generic var in Point would have to conform to Numeric. How can we handle that?

I feel like giving x and y arrays as it is being done now would be simpler for the user
I do not know what Tensor pair is actually, I'll go through the docs.

I've created an issue for this. We can continue the discussion there. #27

PR #25, opened 2019-06-10
Implement scatter plot

KarthikRIyer:
#24 
I've implemented scatter plot. If this seems fine I'll see what other patterns I can implement.
I also plan to add a start and end colour feature and then interpolate between them for all the points.


examples/ScatterPlot/main.swift
```diff
@@ -0,0 +1,36 @@
...
+
+for index in 1...100 {
+    x1.append(Float(index))
+    y1.append(Float(index) + Float.random(in: -10...10))
+}
+
+var agg_renderer: AGGRenderer = AGGRenderer()
+var svg_renderer: SVGRenderer = SVGRenderer()
+
+var plotTitle: PlotTitle = PlotTitle()
```
BradLarson: Unless you need to be explicit about a type, I usually tend to leave the type declaration off of code like this. Also, if all you're going to do later is set the title, I might change this to something like:

```
let plotTitle = PlotTitle(title:"SCATTER PLOT")
```

or even do the initialization inline with setting `scatterPlot.plotTitle = PlotTitle...`.

framework/SwiftPlot/ScatterChart.swift
```diff
@@ -0,0 +1,378 @@
...
+public class ScatterPlot: Plot {
+
+    let MAX_DIV: Float = 50
+
+    let sqrt3: Float = sqrt(3)
+
+    public var xOffset: Float = 0
+    public var yOffset: Float = 0
+
+    public var plotTitle: PlotTitle = PlotTitle()
```
BradLarson: This may apply elsewhere, but thinking about it, would it make sense to have plotTitle, etc. be optionals? That way, they can be selectively disabled when not needed by setting the values to nil.

You could use `if let` logic later to control the drawing or not of each plot feature. This may require a rework of other plots to match, but it would seem to be a clean way of selectively disabling plot elements.

KarthikRIyer:
I've made plot title and plot label optional. I've also removed type declarations from examples wherever applicable.

KarthikRIyer:
I've added color interpolation between start and end color. Interpolation is done according to y values.
Please review.

framework/SwiftPlot/BarChart.swift
```diff
@@ -505,12 +506,16 @@ extension BarGraph {
     }
 
     func drawTitle(renderer: Renderer) {
-        renderer.drawText(text: plotTitle.title, location: plotTitle.titleLocation, textSize: plotTitle.titleSize, strokeWidth: 1.2, angle: 0, isOriginShifted: false)
+        if (plotTitle != nil) {
```
BradLarson: Generally, I try to avoid force-unwrapping where I can, so I tend to use let binding in cases like this. If you use the following:

```
guard let plotTitle = self.plotTitle else { return }
```

followed by your drawing code, you can avoid a level of indentation and force-unwrapping with !. Same with drawLabels and others below.

framework/SwiftPlot/LineChart.swift
```diff
@@ -143,15 +143,16 @@ extension LineGraph{
...
-
-        let xWidth   : Float = renderer.getTextWidth(text: plotLabel.xLabel, textSize: plotLabel.labelSize)
-        let yWidth    : Float = renderer.getTextWidth(text: plotLabel.yLabel, textSize: plotLabel.labelSize)
-        let titleWidth: Float = renderer.getTextWidth(text: plotTitle.title, textSize: plotTitle.titleSize)
-
-        plotLabel.xLabelLocation = Point(((plotBorder.bottomRight.x + plotBorder.bottomLeft.x)/2.0) - xWidth/2.0, plotBorder.bottomLeft.y - plotTitle.titleSize - 0.05*plotDimensions.graphHeight)
-        plotLabel.yLabelLocation = Point((plotBorder.bottomLeft.x - plotTitle.titleSize - 0.05*plotDimensions.graphWidth), ((plotBorder.bottomLeft.y + plotBorder.topLeft.y)/2.0 - yWidth))
-        plotTitle.titleLocation = Point(((plotBorder.topRight.x + plotBorder.topLeft.x)/2.0) - titleWidth/2.0, plotBorder.topLeft.y + plotTitle.titleSize/2.0)
-
+        if (plotLabel != nil) {
```
BradLarson, line 146: This is a case where I'd use an `if let` to bind a value, to avoid bailing early on the plotTitle code later in the function.

BradLarson:
Really like how the use of PlotTitle, etc. initializers cleans up that code. Should make it easy for people to specify those.

I made a few comments above about doing `guard let` or `if let` binding on the optional values instead of the force-unwrapping. While the logic ends up being about the same, I prefer to avoid using force-unwrapping where I can for safety. `guard let` can even save you code and indentation in some of these cases, as well as making it explicit that you want nothing to happen past that point if this value is missing.

Aside from that, this generally looks good to me.

BradLarson (approved):
This looks good to me.

KarthikRIyer:
This isn't done yet. I am adding a few more patterns.

KarthikRIyer:
@BradLarson I've fixed the plot dimensions issue and memory leak issue with AGGRenderer. But I am not entirely sure why what I did worked.

framework/AGGRenderer/CPPAGGRenderer/CPPAGGRenderer.cpp
```diff
@@ -401,7 +427,7 @@ namespace CPPAGGRenderer{
       std::vector<unsigned char> image(buffer, buffer + (frame_width*frame_height*3));
       write_png(image, frame_width, frame_height, file_png);
 
-      // delete[] buffer;
+      delete[] buffer;
```
BradLarson: By deleting the buffer here, will that lead to a problem if someone tries to save the same plot to disk multiple times? The buffer only appears to be initialized on creation of the Plot at a certain size.

KarthikRIyer: @BradLarson yes this might happen if the user draws the graph and tries to save it separately multiple times. I had only thought of the method drawAndSaveGraph() while deleting. 

What if I create a separate method that deletes the buffer and call it in the destructor of each Graph? But this would be AGG specific...

BradLarson: Yeah, somehow you'll want to balance out the new / delete to avoid the leaks you were seeing earlier, while preventing overreleases. It would seem to make sense to delete the buffer on deallocation of the surrounding object, as well as before reshaping the Plot and requiring the creation of a new buffer. Not quite sure where that is, though.

KarthikRIyer: @BradLarson I apologise for the delayed response. I wasn't able to look into it yesterday. I'll get this done and merged today.

KarthikRIyer:
@BradLarson I've deleted the buffer in the deinitializer of the renderer instead of the plot, because I thought that the plot can be rendered only if it has a renderer so in general cases the scope of both of them will be the same. So I think this shouldn't cause any problems. If this seems ok then we can move forward to merge.

BradLarson:
@KarthikRIyer - That seems like it makes sense. Don't know if there's an automated test case that could be set up to test for leaks, but if I could rig this up in a Mac or iOS application I could run Instruments against it and that would provide a clear indication of any leaks. That can come later, though.

When you pull this in, feel free to close my issue about the Mac builds, because this should also take care of that.

Issue #28, opened 2019-06-26
Implement Histogram


KarthikRIyer:

KarthikRIyer:
This is almost done. There are some bugs to sort out in stacked histogram for step type. I hopefully should be able to open a PR for this by tomorrow.

PR #30, opened 2019-06-29
Implement Histogram

KarthikRIyer:
#28 
This is mostly done but I might have to smoothen out a few things.
I had one concern regarding the examples. The ones where I am using the random function. Every time the result will be different. Is it a good idea to store the numbers required for each example in a text file and read them everytime, just so that the examples remain consistent?


BradLarson (approved):
Sorry about the delay, as you may have heard I've been a little busy, also the holidays here slowed me up.

I don't see any fundamental issues with this. When it comes to the random distribution for your test data, you can use [`random(in:using:)` ](https://developer.apple.com/documentation/swift/float/3020582-random) to provide your own RNG where you'd be able to define a constant seed. However, even with that they warn about this changing from Swift version to version.

Might be best to generate values ahead of time, store them in some kind of static array, and use those for your test case.

KarthikRIyer:
Ok. I'm merging this right now. I'll make all the changes in the Pair<> PR.

Issue #32, opened 2019-07-09
Default aesthetics of the plots


BradLarson:
Now that this is getting close enough to the point where people can start using it in notebooks, etc., I think it might help to improve some of the default aesthetics of the plots. The default settings for plots are producing graphs that aren't as aesthetically pleasing as other frameworks, but with a few slight tweaks to fonts, line size, spacing of numbers from the axis hatching, etc. I think the defaults could be made a lot nicer-looking.

For example, as a starting point, it would be nice if the following matplotlib plot:

https://matplotlib.org/gallery/lines_bars_and_markers/simple_plot.html#sphx-glr-gallery-lines-bars-and-markers-simple-plot-py

could be replicated using default settings for the framework, plus enabling a grid. Even that example isn't the greatest in terms of formatting, but people coming from matplotlib will expect something at least that readable.

We should find other examples of really nice-looking plots and see if they can be replicated in this framework with the right settings. That would give good targets for defaults on fonts, etc. at different plot sizes.

KarthikRIyer:
I'll get to work on this next.

KarthikRIyer:
@BradLarson @marcrasi to use a different font with AGG I will need to import truetype as a dependency. AGG has provisions to use different fonts using truetype. So, I tried importing this:  https://github.com/PureSwift/CFreeType

but I am getting this error: 
```console
Fetching https://github.com/PureSwift/CFreeType.git
error: dependency graph is unresolvable; found these conflicting requirements:

Dependencies: 
    https://github.com/PureSwift/CFreeType.git @ 1.0.4
```

The repo has a 1.0.4 tag. Am I doing something wrong?
Meanwhile I'll get to work on #29

BradLarson:
By adding FreeType as a dependency, will that require FreeType to be installed as a library before using the AGG renderer? Or is that commonly installed on systems and in environments like Colab?

What are you using in your Package.swift to pull in CFreeType? That error does sound odd, but I've hit that when I was trying to do an exact match and I got the release numbers slightly wrong.

KarthikRIyer:
FreeType is commonly installed on Ubuntu atleast.

 I think the release number isn't a problem, because I cloned the repo,  deleted the git repository, initialized my own git repository, pushed my own tag, but still got the same error.

BradLarson:
Sorry, mean to ask: what is the exact code you're using in your Package.swift? I can try it on my end with a dummy project to see if I can figure out what's wrong.

[...]

Issue #29, opened 2019-06-27
Cannot plot axis markers with values between 0 and 1


KarthikRIyer:
This has to be solved in all the plots. I'll do this after finishing up histogram, in a separate PR for all the plots.

PR #33, opened 2019-07-14
Improve default aesthetics of plot

KarthikRIyer:
#32 
I've just enabled grid at the moment. I'm opening the PR just to track the progress.


KarthikRIyer:
@BradLarson I have used the default fonts for text in both svg and agg. To support a different font I think I'll have to make significant changes in AGG atleast. I am looking into it now.

KarthikRIyer:
@BradLarson @marcrasi what would be our purpose of rendering SVGs? I mean, why would the end user want to use SVG? Will he/she view it in a browser?
Why I'm asking this is because, on searching, the ways I'm getting, to use custom fonts in an SVG are using CSS like here: https://graphicdesign.stackexchange.com/a/10735
This doesn't work on the default image viewer in Ubuntu, but works in a browser.

BradLarson:
@KarthikRIyer - My primary driver for SVGs was to help make sure the interface was generic enough to support very different kinds of rendering. You don't have access to Core Graphics readily on your end, so this seemed like a good proxy.

I can see a vector backend renderer like this being used for the web as the primary target, so if you want to focus on ways of making that work for custom fonts, that might be a good first step. We can figure out later if something more general is needed for other applications we'd come across.

KarthikRIyer:
@BradLarson @marcrasi I am now able to use custom fonts in both renderers. But it still needs polishing. I am using the default font used in matplotlib i.e. _DejaVuSans_. What I need to figure out next is how to set it as the default font. Now I've hardcoded everything. For AGG the font needs to be present in the directory from where the executable is executed and for SVG the font file needs to be in the same directory as the image.

KarthikRIyer:
@BradLarson @marcrasi I've changed the default font to Roboto-Regular. For SVG images I'm getting the font from the google fonts API. DejaVuSans wasn't available on google fonts, and trying to use the url for that font from anywhere else gave me a CORS error. That's why I decided to go for goggle fonts. So if the end user needs to use a different font he/she needs to pass the name of any font available on google fonts.

For AGG, I'm not sure how I can get a font to be default. I have to pass in the path of the font to the renderer. Right now the font is in the topmost directory of the library, from where we run the command `swift build`. So it will work only when any executable is executed form there. To bundle the font with the library I thought I could use an online tool to convert the ttf file to a  char array and then load it from memory. But AGG doesn't provide any methods to do that. So right now I'm trying to understand AGG's code and will see if I can implement my own method to load the font from memory.

What do you think? Is there a simpler way of achieving this? Or some way we can get the path to the ttf file in the library's directory irrespective of where we execute it from?

KarthikRIyer:
Also I'm getting this:
```console
'CFreeType' /home/karthik/swiftplot/.build/checkouts/CFreeType: warning: system packages are deprecated; use system library targets instead
```

I did as given [here](https://github.com/apple/swift-package-manager/blob/263171977ebcd47f4aaca1202cff5a96c5158a64/Documentation/Usage.md#import-system-libraries) but it's still the same.

Link to CFreeType repo where I made the changes: https://github.com/KarthikRIyer/CFreeType

KarthikRIyer:
#29 
@BradLarson @marcrasi please review. Next I'll move to implementing a CoreGraphics Renderer and simultaneously fix other problems we might notice.

KarthikRIyer:
Also @BradLarson any ideas on how to go about getting it to build on Mac? I believe once FreeType is installed this should build. But I've been unable to install FreeType on Mac.

marcrasi (approved):

KarthikRIyer:
Merging now.

PR #35, opened 2019-08-18
Implement CoreGraphics Renderer code

KarthikRIyer:
#34 
@BradLarson I've initiated the CoreGraphics renderer. For testing I've just used it in the LineChartSingleSeries Example.
Two things are left to implement.
1. Hatching pattern for bar graph(I'll take a look at this by tomorrow)
2. Rotated text to draw y-axis label

Could you please help me out in rotating the text? I'm facing trouble doing it. I am able to find examples in obj-c and when I try to do the same in swift the text gets skewed terribly. The rotation code I tried is in comments.

When this starts working properly we can add the optional compilation according to OS. 
We can also add a function that return an NSImage in case someone wants to display it in an ImageView.

This is only for macOS. 
What kind of an implementation would we require for iOS? It seems like a custom view makes more sense for an iOS implementation, instead of the image files we've been writing till now. Or do we keep everything same as done in the macOS implementation?


KarthikRIyer:
I've implemented hatching pattern and fixed a small offset bug for SubPlots. 
Now the only thing to fix is the rotated text for y-axis label.

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+        var x = p1.x + xOffset
+        if (isOriginShifted) {
+            y = y + (0.1*plotDimensions.subHeight)
+            x = x + (0.1*plotDimensions.subWidth)
+        }
+        let rect = CGRect(x: Double(x),
+                          y: Double(y),
+                          width: Double(w),
+                          height: Double(h))
+        context.setStrokeColor(CGColor(red: CGFloat(strokeColor.r),
```
BradLarson, line 82: I'd recommend creating an extension on Color that returns the equivalent CGColor. It'll save code here and elsewhere, and make things a little easier to read.

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+                          hatchPattern: BarGraphSeriesOptions.Hatching) {
+        switch (hatchPattern.rawValue) {
+        case 0:
+            break
+        case 1:
+            let drawPattern: CGPatternDrawPatternCallback = { _, context in
+                let line = CGMutablePath()
+                line.move(to: CGPoint(x: Double(0), y: Double(0)))
+                line.addLine(to: CGPoint(x: Double(10), y: Double(10)))
+                context.setStrokeColor(NSColor.black.cgColor)
```
BradLarson: NSColor is a Mac-specific type and will cause problems when porting this to iOS. I'd recommend instead to use your platform-independent Color type and the extension I suggested earlier for returning a CGColor value from that.

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+            let destinationURL = URL(fileURLWithPath: fileName)
+            let image = NSImage(cgImage: context.makeImage()!, size: NSZeroSize)
+            if image.pngWrite(to: destinationURL) {
+                print("File Saved")
+            }
+        }
+    }
+}
+
+extension NSImage {
```
BradLarson, line 607: I'd wrap this in an `#if os(macOS)` block to make sure it didn't get built for iOS. Similarly, you could add an `#else` clause that would have the matching code for UIImage on iOS.

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+    var pngData: Data? {
+        guard let tiffRepresentation = tiffRepresentation, let bitmapImage = NSBitmapImageRep(data: tiffRepresentation) else { return nil }
+        return bitmapImage.representation(using: .png, properties: [:])
+    }
+    func pngWrite(to url: URL, options: Data.WritingOptions = .atomic) -> Bool {
```
BradLarson: Would `writePNG(to:options:)` be a better name for this?

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+                         strokeWidth thickness: Float,
+                         angle: Float,
+                         isOriginShifted: Bool){
+        var x1 = p.x + xOffset
+        var y1 = p.y + yOffset
+        if (isOriginShifted) {
+            x1 = x1 + 0.1*plotDimensions.subWidth
+            y1 = y1 + 0.1*plotDimensions.subHeight
+        }
+        let font = NSFont.systemFont(ofSize: CGFloat(size))
```
BradLarson, line 555: NSFont is a another platform-specific type, so you might need to wrap this in an `#if os(macOS` and have a matching path using UIFont for iOS.

framework/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -0,0 +1,599 @@
...
+        let string = NSAttributedString(string: "\(text)", attributes: [NSAttributedString.Key.font: font])
+        let size = string.size()
+        return Float(size.width)
+    }
+
+    public func drawOutput(fileName name: String) {
+        if !name.isEmpty {
+            let fileName = name + ".png"
+            let destinationURL = URL(fileURLWithPath: fileName)
+            let image = NSImage(cgImage: context.makeImage()!, size: NSZeroSize)
```
BradLarson, line 596: You can do an `#if os(macOS)` here to switch between the two image types. If you made the extensions match between them (below), the rest of the code could remain outside of the platform conditional.

KarthikRIyer:
@BradLarson I've made the changes suggested above, but don't know if it builds correctly on iOS.

Also could you please help me out with the text rotation issue? 
Now I'll update the documentation.

KarthikRIyer:
I've update the Package.swift file. Now it builds on macOS and Linux as well.

KarthikRIyer:
Closing. This has been merged along with #42 .

PR #42, opened 2019-11-26
Move examples to become tests

karwa:
This greatly simplifies the Package.swift file, and allows this package to quickly get some test coverage by comparing the test output to the reference files (not done yet - waiting until SwiftPM [gains support for resources](https://forums.swift.org/t/se-0271-package-manager-resources/30730/62) so we can reliably _find_ the reference files).

I've kept the tests as separate files so they can still be easily read as example code, but we could also reasonably group similar examples in to a single file (e.g. all the various hatching patterns).

To run the examples, you now need to run `swift test`. Resulting files will be be placed in `{package-directory}/output/{agg/svg/etc}` if run from the command line, or `/private/tmp/output/{agg/svg/etc}` if run from Xcode.

This requires #35, which has been rebased on top of `master`


karwa:
Oh, and SwiftPM now supports system library targets, obviating the need for the separate `CFreeType` package.

KarthikRIyer:
@karwa this looks really awesome! Thanks a lot for your contribution. There's a small thing that I wanted to fix in QuartzRenderer before merging #35 . I was having trouble rotating text properly. But I guess that can be done after getting this merged considering that GCI starts in a few days and it would be better if such major changes are made before others start contributing.

Tests/swiftplotTests/SwiftPlotTestCase.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+@testable import SwiftPlot
+
+class SwiftPlotTestCase: XCTestCase {
+  // TODO: Possibly allow setting this via a command-line flag?
```
KarthikRIyer: It would be nice if it were possible. I wasn't able to find a way to do this, but the same thing happens when we set the path using --package-path.

karwa:
@KarthikRIyer Sure, I have a need for something like this, and it's a really useful addition to the Swift ecosystem.

I have some more large-ish changes on my fork, including:
- iOS/tvOS/watchOS support. These platforms don't have a system-library package manager like homebrew, so they can't use freetype or the AGG renderer (unless somebody wants to make a Swift package out of freetype). That mostly involves test/example changes, as it means AGG is not available on all platforms as it used to be.
- Adding a `Rect` primitive. Doing so makes a lot of layout code muuuuuuch simpler and easier to read (as opposed to calculating and passing groups of 4 Points around).

If you're happy to merge this, I can open PRs for those.

The next thing I have planned is to refactor some of the layout code - bar-charts, histograms, etc all duplicate lots of code to calculate and draw common things like the plot border, title and legend. Putting that all in one place will make it easier to improve, or to use in new kinds of charts.

I'll also take a look at the CG text rotation. Another thing I'm interested in with the CG renderer is drawing on an externally-provided context (e.g. the drawing context of a UIView or a PDF context).

KarthikRIyer:
Merging.

These changes you've planned seem really great! But I don't understand why these platforms won't be able to use AGG. It's bundled with swiftplot. As far as text drawing is concerned AGG can draw text without freetype but that default font isn't good 😅.

karwa:
@KarthikRIyer Maybe it is possible, but I couldn't find a simple way to disable it (SwiftPM only supports platform-specific compiler flags right now, so we'd need a `#define` to disable freetype)

PR #44, opened 2019-11-27
[feat] Expand QuartzRenderer to support iOS/tvOS/watchOS

karwa:
Tested this by running the examples in the iOS and tvOS simulators. You need to comment out AGG-related stuff as explained in the Package.swift file.

I also renamed the source directory to `Sources`, if you don't mind. It's not a big deal, but that's kind of the idiomatic place where all Swift packages tend to put them, and SwiftPM will automatically find them if they're in there (notice we now don't need to specify the `path` for most targets in the `Package.swift` file. We only need to do it for the CPP/C-Agg layers because they're grouped in the same subfolder).

The Package.swift file is already a bit busy because of the platform-specific parts, and moving the files like this allows us to reduce it a little bit.


KarthikRIyer:
Merging. Thanks!

PR #45, opened 2019-11-28
Add `Rect` and `Size` primitives and use them for layout

karwa:
This massively simplifies lots of layout maths (especially in BarChart, as you might expect).

I'm currently working on tweaking the layout so that x/y axis labels don't intersect with markers, and stuff like that is made much easier with a `Rect` type.

### DIFF
--- a/Sources/SwiftPlot/Pair.swift
+++ b/Sources/SwiftPlot/Pair.swift
@@ -15,9 +15,68 @@ public let zeroPoint = Point(0.0, 0.0)
 public struct Size {
     public var width: Float
     public var height: Float
+    
+    public init(width: Float, height: Float) {
+        self.width = width
+        self.height = height
+    }
+}
+extension Size {
+    public static let zero = Size(width: 0, height: 0)
 }
 
+/// A Rectangle in a bottom-left coordinate space.
+///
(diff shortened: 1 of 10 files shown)


Sources/SwiftPlot/Pair.swift
```diff
@@ -11,3 +11,13 @@ public struct Pair<T,U> {
...
+
+public struct Size {
+    public var width: Float
+    public var height: Float
+}
+
+public struct Rect {
+    public var origin: Point
+    public var size: Size
+}
```
KarthikRIyer: We could have separate files for these for clarity, or maybe rename this file to Primitives.swift?

KarthikRIyer:
The rest of it looks realy good.

karwa:
@KarthikRIyer Yeah I think we could do with a general set of geometry types. For polar plots for example, I imagine we might like a `Circle` type with helpers for working with circles.

Is it okay to do that in a separate PR? I have another big one after this to unify the layout code which I'd like to land first.

KarthikRIyer:
@karwa sure!

KarthikRIyer:
Merging.

PR #46, opened 2019-11-28
Unify graph plot layout

karwa:
I noticed that all the plots we have currently do a lot of work to set up their plot chrome. By unifying them, we can achieve greater consistency, fewer bugs, and it will be easier to make improvements. For example, I'm currently looking at a way to tweak the layout so that axis labels will not overlap the nearby markers. By doing this on GraphLayout, all charts automatically benefit.

The way it works is to move the layout calculation/drawing logic and the data which affects it to a separate struct. The data properties are exposed on the object via the `HasGraphLayout` protocol, so code like `barChart.plotTitle = ...` still works, even though the storage has moved to the layout object.

The great thing about this is that it makes the code super-easy: the charts now only need to worry about formatting (scaling) and drawing their data in their unique way. And it's super-easy to make a new chart and get some chrome:

```swift
class MyGraph: Plot, HasGraphLayout {
            
            var layout: GraphLayout
            var xOffset: Float = 0
            var yOffset: Float = 0
            
            init(dimensions: PlotDimensions = .init(frameWidth: 1000, frameHeight: 700)) {
                layout = GraphLayout(plotDimensions: dimensions)
            }
            
            func calculateScaleAndMarkerLocations(markers: inout PlotMarkers, renderer: Renderer) {
                
                markers.xMarkers = Array(stride(from: 0, to: 500, by: 100))
                markers.xMarkersText = markers.xMarkers.map { String($0) }
                
                markers.yMarkers = Array(stride(from: 0, to: 500, by: 100))
                markers.yMarkersText = markers.yMarkers.map { String($0) }
                
                markers.y2Markers = Array(stride(from: 0, to: 500, by: 150))
                markers.y2MarkersText = markers.y2Markers.map { String($0) }
            }
            
            func drawData(markers: PlotMarkers, renderer: Renderer) {
            }
            
            var legendLabels: [(String, LegendIcon)] { [
                ("Something", .square(.green)),
                ("Another thing", .shape(.star, .lightBlue)),
            ]}
        }
```

Also added support for text colours. I would have liked to separate that out to keep the PR as easy to read as possible, but it ended up being too tricky to cut it out.


Sources/SVGRenderer/SVGRenderer.swift
```diff
@@ -180,7 +180,7 @@ public class SVGRenderer: Renderer{
             y = y + 0.1*plotDimensions.subHeight
         }
         y = plotDimensions.subHeight - y
-        let circle: String = "<circle cx=\"\(x)\" cy=\"\(y)\" r=\"\(r)\"  style=\"fill:rgb(\(fillColor.r*255.0),\(fillColor.g*255.0),\(fillColor.b*255.0));opacity:\(fillColor.a)\" />"
+        let circle: String = "<circle cx=\"\(x)\" cy=\"\(y)\" r=\"\(r)\"  style=\"fill:(\(fillColor.svgColorString);opacity:\(fillColor.a)\" />"
```
KarthikRIyer: let circle: String = "<circle cx=\"\(x)\" cy=\"\(y)\" r=\"\(r)\"  style=\"fill:\(fillColor.svgColorString);opacity:\(fillColor.a)\" />"

KarthikRIyer (changes requested):
Circles are getting rendered black right now. The bracket before fillColor.svgString seems to be the reason. The necessary bracket is included in svgString and needs to be removed from here.

Sources/SVGRenderer/SVGRenderer.swift
karwa: Thanks, I've pushed an update

KarthikRIyer:
Thanks! Merging.

PR #47, opened 2019-11-30
Fix colours, angled drawing on CoreGraphics

karwa:
Apparently nobody noticed that when the example said `.orange`, it was actually more of a mustard  colour 😅


KarthikRIyer:
Oh! Never noticed the color bug.😂 
And thanks for the rotation in quartz! I wonder why I wasn't able to figure out how to do that.

Merging.

PR #48, opened 2019-11-30
New layout algorithm for GraphLayout

karwa:
- Now that graph layout is unified, we can make improvements to all of them at once.
- Includes work to make the graph drawing more relocatable.
- Makes saving a file be a throwing operation. For example, if you try to save to a subfolder which doesn't exist, that would just fail with some debug output. Now you get an error, which you can ignore with `try?` if you don't care about.
- The SVG renderer is a good candidate for Swift's [raw string literals](https://github.com/apple/swift-evolution/blob/master/proposals/0200-raw-string-escaping.md)

After this, the only stuff I've got left is setting up QuartzRenderer to use an external context and hooking that up to AppKit, UIKit and SwiftUI. I have a [proof-of-concept version](https://github.com/karwa/swiftplotviewtest) working.


KarthikRIyer:
This looks really good! This should definitely make it easier to work on this project further. Thanks!

Merging.

KarthikRIyer:
@karwa could you please also update the docs wherever required. I think most of the changes will be in the property tables.

karwa:
Sure, will do.

KarthikRIyer:
@karwa `lineGraph.drawGraph(renderer: agg_renderer)` this doesn't seem to work. It's asking for the size parameter but I can see that there's a function with a default size.

karwa:
@KarthikRIyer ah whoops - apparently I forgot to make the one with the default size `public` 😅

The examples/tests all use `drawGraphAndOutput`, so I didn't notice. We should definitely add some real tests which exercise more of the API.

KarthikRIyer:
😄 I didn't notice it wasn't public. I've been trying to figure out why that was happening for the past few minutes. Yeah we should make the tests cover more of the API.

PR #49, opened 2019-11-30
Make charts more position-independent, add support for external CG contexts

karwa:
OK this is all I've got, now 😄

- Adds support for external CG contexts (e.g. from AppKit/UIKit views)
- Replaces clamping of line-chart functions with a clamping parameter... which doesn't really "clamp", instead it "clips". Naming suggestions welcome.
- Make charts more size-independent. The idea of them having a specific size was always a bit of a lie, because you could put them in a subplot and their size would change. Really, the renderer determines the size. Anyway, allowing this enables us to...
- Make SubPlot a Plot! Together with nested SubPlots, this also unifies the API for drawing plots and subplots (which is helpful for wrappers like native views)

And I'll update the documentation next and add an example of a nested subplot.


Sources/QuartzRenderer/QuartzRenderer.swift
```diff
@@ -56,21 +56,13 @@ public class QuartzRenderer: Renderer {
                                  space: Self.colorSpace,
                                  bitmapInfo: Self.bitmapInfo)!
         self.isExternalContext = false
-        
-        let rect = CGRect(origin: .zero, size: CGSize(imageSize))
-        context.setFillColor(Color.white.cgColor)
```
KarthikRIyer: Isn't a white background a better default setting? Both SVG and AGG backends have a default whie background.

KarthikRIyer (commented):
I really like the idea of making subplots recursive. Can we have an example for it?
Also an example for using QuartzRenderer with AppKit/UIKit views?

Sources/QuartzRenderer/QuartzRenderer.swift
karwa: The GraphLayout already fills the area in a background color, so it isn't necessary to do it again.

karwa:
@KarthikRIyer I've added an examples of a nested subplot and updated the README examples. I think we should just remove the "Documentation" section and replace it with actual `///` doc-comments.

Swift has an [official doc-comment style](https://nshipster.com/swift-documentation/) which I believe is understood by the compiler and may even be exposed via SourceKit (Xcode certainly knows about them and renders them. Not sure if VSCode/SourceKit can do that).

I'll push the view adapters afterwards. One thing I'm working on is that, because the charts are classes, anything that contains a chart (like a view adapter) doesn't know when the chart's data changes - only if you totally replace it with a new chart. My fix for that is very drastic but also super-cool: we can make all the charts value-types! As an additional benefit, that means every chart is able to be rendered on background threads and the API in general becomes much simpler:

```swift
var lineGraph1 = LineGraph<Float, Float>(...)
lineGraph1.addSeries(x, y, label: "Plot 1", color: .lightBlue)
// blah, blah, set up titles, labels, etc...

var lineGraph2 = lineGraph1 // This is now a fully-independent chart, not another reference to lineGraph1.
lineGraph2.addSeries(x, y, label: "Plot 1", color: .lightBlue)

let subplots = Subplots(layout: .horizontal, plots: [lineGraph1, lineGraph2])
```

KarthikRIyer:
@karwa thanks a lot for adding the examples. I wasn't aware of the official doc-comments. I think doing that properly will take some time. As [Google-Code In](https://codein.withgoogle.com/) starts today and this project is a part of it, could we shift to the new doc style after the contest gets over? Just so that a major change doesn't cause inconvenience to the participants during the contest.

KarthikRIyer:
Merging

karwa:
Oh sure. I never heard of Google Code-in.

PR #52, opened 2019-12-04
Change linecap style for consistency and fix solid lines that should be dashed

odmir:
This makes SVG rendering more consistent with the other renderers by using the `butt` linecap style instead of `round` and changes how consecutive lines are drawn by using the `<polyline>` SVG element.
These changes also fix consecutive line segments being visibly separated/broken and the dashed plots now actually appears with a dashed style instead of being solid. Another benefit is that SVG file sizes can now be much smaller and rendering is possibly faster (the file '_06_function_plot_line_chart.svg' went from 73 KB down to 19 KB and '_25_nested_subplots.svg' went from 255 KB down to 124 KB for example).

Before & after:
Screenshot 2019-12-04 at 00 48 58
Screenshot 2019-12-03 at 23 34 49
Screenshot 2019-12-03 at 23 35 21

Before & after:
Screenshot 2019-12-04 at 00 51 55
Screenshot 2019-12-04 at 00 53 36


KarthikRIyer:
Thanks for the PR! Appreciated.
This looks good to me. Merging.

Sources/SVGRenderer/SVGRenderer.swift
```diff
@@ -189,10 +189,16 @@ public class SVGRenderer: Renderer{
                               strokeWidth thickness: Float,
                               strokeColor: Color,
                               isDashed: Bool) {
-        guard !p.isEmpty else { return }
-        for i in 0..<p.count-1 {
-            drawLine(startPoint: p[i], endPoint: p[i+1], strokeWidth: thickness, strokeColor: strokeColor, isDashed: isDashed)
-        }
+        guard p.count > 1 else { return }
+        
+        let pointsString = p.map { point in
```
karwa, line 194: `p.map` creates an Array, which we then immediately throw away after joining the elements together in to a String. It's _possible_ that the optimiser could remove the temporary, but not guaranteed or even predictable. `p.lazy.map` would always avoid the extra allocation, however.

It's not a massive issue, but I thought you might be interested.

Sources/SVGRenderer/SVGRenderer.swift
```diff
@@ -189,10 +189,16 @@ public class SVGRenderer: Renderer{
...
+            let convertedPoint = convertToSVGCoordinates(point)
+            return "\(convertedPoint.x),\(convertedPoint.y)"
+        }.joined(separator: " ")
+        
+        let dashedString = isDashed ? ";stroke-dasharray:4 1" : ""
```
karwa, line 199: Would you mind moving the leading semicolon to the end, and restoring the semicolon after the fill-style? So it would become:

```swift
let dashedString = isDashed ? "stroke-dasharray:4 1;" : ""

// ...
fill:none;\#(dashedString)" />"#)
```

This would help ensure that reordering or adding to the interpolations doesn't break the string, but in general it's just strange to see a separator at the start of a segment.

KarthikRIyer:
@karwa thanks for the suggestions. They were really helpful.

@odmir could you please make the changes suggested above in a separate PR?

odmir:
Sure, will do. Also thank you for the suggestions, always learning!

PR #54, opened 2019-12-04
Make Plot.drawGraph(Renderer) public, faster base64 encoding

karwa:
Some small tweaks:

- Make Plot.drawGraph(Renderer) public
- Faster base64 encoding using a non-copying Data object
- Added a rudimentary test for base64 encoding


KarthikRIyer:
The new test hasn't yet been added to __allTests() yet, so it doesn't run when using swift test. The same with the test for SubPlots.

karwa:
@KarthikRIyer done. SwiftPM is actually able to discover tests if you give it the `--enable-test-discovery` flag, but for some reason that's not the default ¯\_(ツ)_/¯

KarthikRIyer:
Thanks @karwa . There seems to be some talk about making it default [here](https://forums.swift.org/t/make-test-discovery-on-by-default/30321) but nothing conclusive.

Merging.

PR #55, opened 2019-12-04
Darwin/CG fixes

karwa:
Also, made it easier to build an iOS/tvOS/watchOS compatible library (all you need to do is change 1 line in Package.swift) and simplified a BarChart method.

Splitting these off because I have some bigger things I want to land after GCI is done.

Screenshot 2019-11-30 at 23 52 52

...and yes, I really did test plotting a chart on an Apple Watch. An animated one, at that 😎

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -45,6 +45,8 @@ let platformProducts: [Product] =  [
   .library(name: "AGGRenderer", targets: ["AGGRenderer"]),
 ]
 
+// Building for iOS/tvOS/watchOS:
+// Change the condition below to "!os(macOS)".
 #elseif os(macOS)
 
 let platformTargets: [Target] = [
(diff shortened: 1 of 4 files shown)


KarthikRIyer:
This is really cool! 😄 
For animation do you simply update the points or is there something more to it?

karwa:
@KarthikRIyer I know, right? I was really excited when it started working. Obviously our defaults for things like padding, font sizes and markers would need adapting to the smaller size, but smartwatches are being used for various tracking and monitoring tasks where charts are really useful - so it could be an interesting use-case.

For now the animation is just updating points and redrawing (although it's happening automatically via SwiftUI's `@State` variables, which is cool). There may be opportunities to optimise that - maybe doing the background separately and not redrawing if it didn't change.

One of the nice things about moving to the `GraphLayout` structure is that it has the entire geometry of the chart in its `Results` structure. So if we make that `Equatable`, we can check it against the previous value and just not redraw the background (almost -- there are non-layout changes which still require a redraw, like colour changes).

The bigger patches I want to land move the `scaledValues` from `Series` and in to a data-structure which the charts create in their `calculate` method and get passed back in their `drawData` method. That means that laying-out and drawing a chart doesn't mutate the chart's state in any way. I'm hoping that can eventually allow charts to be rendered on a background thread (smartwatches today are nuts - multi-core, 64-bit, all that. Totally bonkers). 

The only remaining hurdle for background rendering is that the `Series` still stores colours, etc. Perhaps that can be replaced with some kind of stylesheet structure ([like Matplotlib](https://matplotlib.org/gallery/index.html#style-sheets)) which could also be passed through the pipeline, but I haven't done anything with that yet.

PR #58, opened 2019-12-04
Fix histograms

### LINKED ISSUE #57: Misaligned Histogram Output with CoreGraphics
I was doing some testing with histograms and found out that for some reason the CoreGraphics renderer has a weird y- and x-offset bug, compare:

SVG output:
Screenshot 2019-12-04 at 18 54 19

CoreGraphics output:
_24_histogram_stacked_step_line_joins

I just created a histogram with 10 bins with one series with the element 5 and another series with the element 6 in it. Do we know why this is?

karwa:
We were failing to clear some state in the histogram, leading to errors the second/third/etc time it was rendered.

Fixes #57.


KarthikRIyer:
Merging.

PR #59, opened 2019-12-05
Verify that test output files actually match the reference files

karwa:
Also attempted to improve the base64 encoding test - and I think I may have found a bug (perhaps related to the Jupyter issues?)

This new version of the test copies the logic from the cross-hatched barchart test, draws it directly in to a base64 string (i.e. in-memory), and compares it to the base64-encoded disk output. I think they should be the same, but I'm not 100% sure (does the disk version have extra metadata or something?) 

We know from the original BarChart test that the AGG renderer produces identical binary files when writing to disk - however, the test I described above fails (i.e. in-memory base64 string is different).

Moreover, the in-memory output appears invalid and many browsers do not seem to want to render it. The base64 string from the disk version was rendered correctly.


karwa:
Hmm 😕 Well, hey - a couple of failures means that the test is probably actually testing something! 😅

KarthikRIyer:
@karwa I tried plotting in Jupyter (not google colab). The example line plot worked the first time. But failed when I tried to execute the cell with plotting code again. I don't know why the image encoded in memory and saved to disk should be different. Could it be a bug in lodepng? Maybe i could try using another library? What do you suggest?

Also now there are some other tests failing (all seem to be svg), most of them are line chart tests.

KarthikRIyer:
@karwa the comments in lodepng.h do not say anything about extra metadata while writing to disk.

karwa:
@KarthikRIyer Of course, the recent minor edit added a semi-colon, which matter in an SVG text file...

I'll look in to the PNG stuff. At least we have a test, which should pass but fails due to this issue. As I said, I pasted the output base64 PNG string in to an online viewer, and it wouldn't take it. The base64 from disk was accepted, so I think we probably are doing something wrong. We just need to find out _what_ 😅

PR #60, opened 2019-12-05
Fix base64 PNG rendering

karwa:
`lodepng_encode_memory` gives us a buffer which is (according to the now-working test) indistinguishable from the file on disk.

I also removed the base64 newlines - it makes it harder to debug, because lots of tools don't them (besides, they increase the size, as the test also shows). Are they required for Colab/Jupyter?

Colab still doesn't display the graphs (maybe Jupyter does?), but it doesn't look like our problem; I pasted the base64 string in to https://codebeautify.org/base64-to-image-converter (basically, what came up when I googled "base64 png viewer"), and it displays just fine.

Also fixed leaking plots when setting image size, and leaking non-buffer plot memory in general.

### DIFF
--- a/Sources/AGGRenderer/AGGRenderer/AGGRenderer.swift
+++ b/Sources/AGGRenderer/AGGRenderer/AGGRenderer.swift
@@ -7,10 +7,11 @@ public class AGGRenderer: Renderer{
     public var offset = zeroPoint
     public var imageSize: Size {
         willSet {
-            agg_object = initializePlot(newValue.width, newValue.height, fontPath)
+          delete_plot(agg_object);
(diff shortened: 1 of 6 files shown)


KarthikRIyer:
Oh, have I been using the encode function? I might've missed that. 😢 It worked on one example at that time. So I thought I was using the correct function. Even when I was looking at the comments yesterday I was looking at the encode_memory function thinking that was what I'd used. Thanks for debugging!

KarthikRIyer:
No the newlines aren't a requirement for jupyter/colab.

KarthikRIyer:
@karwa yes, it works on Jupyter, but not on colab.

KarthikRIyer:
Merging

PR #61, opened 2019-12-07
docs: update dependency addition

WilliamHYZhang:
The 1.0.0 release has a separate dependency on CFreeType that seems to have a problem. A recent PR moved that dependency into the package itself, as SwiftPM now supports system library targets. This solved the issue. SwiftPlot uses Anti-Grain Geometry (AGG), a C++ library as a rendering backend, which in turn uses freetype to render text.


KarthikRIyer:
Thanks @WilliamHYZhang! Appreciated.

PR #65, opened 2019-12-15
feat: add annotation support

WilliamHYZhang:
Hi,

This is my PR for adding the watermark/annotation features in the graph plots. 

I tried to mimic the style of the ```plotTitle``` attribute as they were very similar in function.

I'm a bit confused on what @karwa meant by type ```[Annotation]```. Would that be an argument in the ```drawAnnotation``` function in line 210 of ```GraphLayout.swift```? Would I also have to define a structure of that type somehow? Currently, I just have a member ```plotAnnotation``` that contains the annotation, color of text, size, etc.

Running ```swift build``` on the repo returns without any error, however when I was testing using this function in another directory with my fork of swiftplot in the dependencies, when I set the annotation using ```lineGraph.plotAnnotation.annotation = "HELLO WORLD"```, running ```swift run``` returns an error saying ```error: value of type 'LineGraph<Float, Float>' has no member 'plotAnnotation'```. However, I believe that I have defined the member ```plotAnnotation``` in line 15 of ```GraphLayout.swift```. Is there something that I am missing here?

Thanks so much for your help and understanding, I'm completely new to Swift and its syntax.

### DIFF
--- a/Sources/SwiftPlot/GraphLayout.swift
+++ b/Sources/SwiftPlot/GraphLayout.swift
@@ -12,6 +12,7 @@ public struct GraphLayout {
     var backgroundColor: Color = .white
     var plotBackgroundColor: Color?
     var plotTitle = PlotTitle()
+    var plotAnnotation = PlotAnnotation()
     var plotLabel  = PlotLabel()
     var plotLegend = PlotLegend()
     var plotBorder = PlotBorder()
@@ -39,6 +40,7 @@ public struct GraphLayout {
         var yLabelLocation: Point?
         var y2LabelLocation: Point?
         var titleLocation: Point?
+        var annotationLocation: Point?
         
         var plotMarkers = PlotMarkers()
         var xMarkersTextLocation = [Point]()
@@ -200,6 +202,17 @@ public struct GraphLayout {
                               location: titleLocation,
                               textSize: plotTitle.size,
                               color: plotTitle.color,
+                              strokeWidth: 1.2,
+                              angle: 0)
+        }
+    }
+
+    func drawAnnotation(results: Results, renderer: Renderer) {
+        if let annotationLocation = results.annotationLocation {
+            renderer.drawText(text: plotAnnotation.annotation,
+                              location: annotationLocation,
+                              textSize: plotAnnotation.size,
+                              color: plotAnnotation.color,
                               strokeWidth: 1.2,
                               angle: 0)
         }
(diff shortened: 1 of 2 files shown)


KarthikRIyer:
@WilliamHYZhang first of all, good job, considering that this is the first time you're using Swift!

I'm not sure why you're getting that error. I'm testing using the examples available in the Tests folder and it works.

What @karwa  probably meant by Annotation was to create a [protocol](https://docs.swift.org/swift-book/LanguageGuide/Protocols.html) which define some common properties necessary for any type of Annotation. If you take a look at [matplotlib annotations](https://matplotlib.org/3.1.1/gallery/index.html#text-labels-and-annotations) or just google for matplotlib annotations you'll find that they are some additional things like overlay text, arrows, boxes, etc that one can add to a plot. Some of the common properties of an annotation might be it's position, size, color, etc. So, the Annotation protocol will make sure that any type of annotation that we make have these properties, and of course they can have other properties too.

You should create such a protocol, create a Text Annotation type that conforms to the Annotation protocol and move the annotation location inside the Text Annotation type. You also need to set the location of the watermark before drawing.

You could maybe have an array of annotations in **HasGraphLayout** and define a function **addAnnotation(annotation: Annotation)** that accepts an Annotation type which appends the passed in annotation to the array. And the **drawAnnotation** function could iterate through the array and draw all of them. Another thing comes to mind, if we have other kinds of annotations like boxes and arrows each of them will be draw differently, so the protocol should also have a **drawAnnotation** function that takes in a renderer. And while iterating through the annotation array you just call this function for each element and pass in the renderer.

So in the end all that the user would have to do will be something like this:

var lineGraph = LineGraph<Float, Float>(enablePrimaryGridAxis: true)
...
```swift
lineGraph.addAnnotation(annotation: TextAnnotation(text:"HELLO WORLD", location: Point(0.95, 0.95), color: .black))
```
I hope this makes it clear the flexibility we gain by defining a protocol.  If it still seems a bit confusing, read a bit about protocols from the link above or you should be able to find videos on YouTube as well, and take some time to see how the Plot protocol works. You can see this in Plot.swift and any plot type of your choice. Hope this helps and if you have any more questions feel free to ask!

WilliamHYZhang:
@KarthikRIyer thank you so much for the detailed help! Appreciate it a lot. I will read up on protocols today and try to implement one for annotation. I'll push it onto my fork when I'm done so you can see my progress.

WilliamHYZhang:
Update: I'll probably have to rewrite all the code for this, so I'll close this PR and open a new one when I'm done. Thanks!

Issue #40, opened 2019-11-25
Allow adding a text watermark to all the plots


KarthikRIyer:
This task will require you to implement the following:

- Add a new feature to existing plots that let the end users specify some text and use it as a watermark for their plot. The text drawing functions already exist. You need to add a function to the plot types that accept a String and use the drawing functions to add a watermark.

WilliamHYZhang:
Hi, I'm planning on implementing this soon. Thanks!

WilliamHYZhang:
Hi, thanks so much for your kind help before getting set me up with swiftplot on the Google Group. I was looking through the source files and was wondering where I should put the code to add the feature. I see the drawText function in /Sources/SwiftPlot/Renderer.swift, however, I don't know where the "plot types" file is.

KarthikRIyer:
@WilliamHYZhang The Plot Types are in the SwiftPlot directory, for example BarChart.swift, LineChart.swift, etc. You do not need to add the function to draw text to each of these plots. Take a look at BarChart.swift for example and try to understand how it works. As a hint, you would need to make changes to GraphLayout.swift, which is the file that contains the code to draw stuff that's common to all the plots. Hope that helps.

Also, you should claim the task on the GCI website, and also state which issue you are working on before proceeding to work on the task.

Best of luck! I look forward to seeing your work! 😄

karwa:
@WilliamHYZhang Just some implementation advice here:

If you take [a look at Matplotlib](https://matplotlib.org/3.1.1/gallery/index.html#text-labels-and-annotations), you'll see that [watermarks and text labels](https://matplotlib.org/3.1.1/gallery/text_labels_and_annotations/watermark_text.html#sphx-glr-gallery-text-labels-and-annotations-watermark-text-py) are just one of the kind of annotations they support.

So I would recommend making something like an `Annotation` protocol. Very simple, like `Plot`, just describing some object which can be drawn in a Rect using a given Renderer. Then you'd add a property on `GraphLayout` of type `[Annotation]`, where the user can add labels or anything else we add later (arrows, etc). You can expose this on the plots themselves with `HasGraphLayout`, like we do for `plotTitle`.

Then you'd add a label type which conforms to `Annotation`, and make `GraphLayout` tell them to draw when it draws the rest of the foreground.

WilliamHYZhang:
[...]

PR #66, opened 2019-12-15
Add Annotation Support

WilliamHYZhang:
#40 
Hi,

This is my new PR for the addition of annotation support following the advice and help of @KarthikRIyer and @karwa. 

First, I have created a protocol of type ```Annotation``` in ```PlotStyleHelpers.swift``` that contains common properties such as the size, color, and location, and also a ```TextAnnotation``` of type ```Annotation``` that contains the text (specific to TextAnnotation) and the aforementioned properties in ```Annotation```. Please let me know if that was the correct location to put those definitions, thanks!

Next, I updated the ```GraphLayout.swift``` file. I added an array of type ```Annotation```, initialized as an empty array. Annotations would be added to this array with the ```addAnotation``` mutable function that accepts type ```Annotation```. So, if you want to add an annotation you would call:
```
lineGraph.addAnnotation(annotation: TextAnnotation(text:"HELLO WORLD", location: Point(0.95, 0.95), color: .black))
```
as @KarthikRIyer demonstrated in my previous PR.

I also defined a function, ```drawAnnotations```, that is called when ```drawForeground``` is called. This function takes in the renderer and iterates through all of the annotations. To make this function compatible with all types of Annotations (currently now it's only TextAnnotation, but going forward annotations such as boxes, arrows, etc.), I have this function check if it is of type ```xAnnotation``` (as this stage just ```TextAnnotation```, and if it is I perform an explicit downcast and perform the drawing function for that specific annotation. So, if in the future we want more types of annotations, we would simply define a new type conforming to the protocol ```Annotation```, and add an if statement in the ```drawAnnotations``` function to handle that type.

Please let me know what you think!

P.S. @KarthikRIyer In your previous comment you suggested to put ```drawAnnotation``` function in the protocol, however would it make more sense to place it in GraphLayout as I did in this PR, as that is where the other draw functions are put? Thanks for your feedback. 

EDIT:
Now, the ```drawAnnotation``` function is a method of ```Annotation``` and is called in the ```drawAnnotations``` function of ```GraphLayout.swift``` for each annotation. Thanks to @odmir for his comment.


odmir:
Hi! This looks really nice!

I think that what @KarthikRIyer meant is that the `Annotation` protocol should have a `drawAnnotation` method requirement. Then types conforming to that protocol, like `TextAnnotation`, will be required to implement that method. This is a nice thing, new types that want to "be" an `Annotation` will have to have a `drawAnnotation` method or the compiler will complain. This is to say, the annotations should know how to draw themselves and the protocol makes sure that they do. Then the `drawAnnotations` method of the `GraphLayout` struct should only really need to iterate over all `Annotation`s in the array and directly call `drawAnnotation` on them (passing the renderer as argument), without needing to test if it is a `TextAnnotation` or an `ArrowAnnotation` first.

When eventually a new kind of annotation comes along it only needs to conform to `Annotation` and implement a `drawAnnotation` method. You will not need to edit the `drawAnnotations` method on `GraphLayout` and it will just work.

Hope this helps!

WilliamHYZhang:
@odmir I see what @KarthikRIyer means now, thanks so much for clearing it up! I'll fix that up and push a new commit when I'm done. Thank you again!

Sources/SwiftPlot/PlotStyleHelpers.swift
```diff
@@ -42,3 +42,17 @@ public struct PlotLegend {
...
+    var location: Point {get set }
+}
+
+struct TextAnnotation : Annotation {
+    public var text = ""
+    public var color = Color.black
+    public var size: Float = 15
+    public var location = Point(0.0, 0.0)
+    public init() {}
+}
```
odmir: Explicitly defining an init tells Swift not to synthesise one automatically for the stored properties. An empty init makes it so the only way to create a text annotation is to do `TextAnnotation()`. It will not be possible to specify any property in the initialiser and every `TextAnnotation` created will always have the initial values specified here. If you remove `public init() {}` Swift will synthesise two initialisers automatically: `TextAnnotation()` (this one is synthesised since there are default values for all the properties) and `TextAnnotation(text: String, color: Color, size: Float, location: Point)`.

WilliamHYZhang: Ok, I will remove that now. Thank you for your review, appreciate it a lot, especially since I'm getting accustomed to Swift syntax :)

KarthikRIyer: @odmir @WilliamHYZhang although not specifying an init will generate one automatically, its access level will be internal. You'll need to specify an init like so to be able to initialize annotations outside PlotStyleHelpers.swift:

```swift
public init(text: String = "", color: Color = .black, size: Float = 15, location: Point = Point(0.0, 0.0)) {
      self.text = text
      self.color = color
      self.size = size
      self.location = location
    }
```

Sources/SwiftPlot/GraphLayout.swift
```diff
@@ -363,6 +365,21 @@ public struct GraphLayout {
                               angle: 0)
         }
     }
+
+    mutating func addAnnotation(annotation: Annotation) {
```
KarthikRIyer: @WilliamHYZhang you have declared this function in GraphLayout so it will be inaccessible to any plot. Move it to 
`extension Plot where Self: HasGraphLayout {` .
This is an extension for `Plot` so each plot will have access to this function.

Sources/SwiftPlot/GraphLayout.swift
```diff
@@ -383,6 +400,10 @@ extension HasGraphLayout {
         return []
     }
     
+    public var annotations: [Annotation] {
+        return []
+    }
+
```
KarthikRIyer: This makes the annotations a read only variable as you have no setter. Instead you should have:
```swift
public var annotations: [Annotation] {
        get { layout.annotations }
        set { layout.annotations = newValue }
    }
```

Sources/SwiftPlot/PlotStyleHelpers.swift
```diff
@@ -42,3 +42,18 @@ public struct PlotLegend {
     public var textSize: Float = 10
     public init() {}
 }
+
+public protocol Annotation {
+    var text: String { get set }
+    var color: Color { get set }
+    var size: Float { get set }
```
KarthikRIyer: Instead of `var size: Float { get set }` I'd like it to be of the type `Size` which you can find in `Pair.swift`. This will be useful when you need to draw a rect or an arrow. Having a float size doesn't make sense for these applications. You can still have a `var textSize: Float` in `TextAnnotation`. In `TextAnnotation` the size can be a [computed property](https://medium.com/@micosmin/swift-computed-properties-2a8fd9f88d6). You will be able to get the bounding box/Size for a text from any renderer. Check Line 184 in Renderer.swift.

Sources/SwiftPlot/PlotStyleHelpers.swift
```diff
@@ -42,3 +42,18 @@ public struct PlotLegend {
     public var textSize: Float = 10
     public init() {}
 }
+
+public protocol Annotation {
+    var text: String { get set }
```
KarthikRIyer: You can remove this from Annotation protocol. It won't make sense to have a text in an ArrowAnnotation which we might create in the future.

KarthikRIyer (changes requested):

KarthikRIyer:
@WilliamHYZhang in addition to the above comments you need to remove the PlotAnnotation type you created earlier.

Also, have you tested your code on an example? You can edit the existing tests in the `Tests` folder temporarily to see if your feature works as expected, and then when you're done with it you should make a new test for this feature and update the reference images in `Tests/Reference`. You can run the tests simply bu running `swift test`.

In the watermark test It'd be nice if the watermark was a bit transparent so instead of using `.black` as the color you could use Color(0.0, 0.0, 0.0, 0.5).

Take a look at `XCTestManifests.swift` to see how the tests are made.

WilliamHYZhang:
Thank you for your feedback! I'll work on implementing these changes today.

karwa:
A couple of points:

- I would strip the `Annotation` protocol down to only the drawing function.

  The protocol is only useful for the `GraphLayout`, which itself doesn't care about the annotation's colour, or even its size or location. The only thing the `GraphLayout` cares about is that the annotation can draw itself with the renderer that we give it.

  And the function could just be named `draw`. We know it's an annotation - there's no need to say it again by calling the method `drawAnnotation`.

- I wonder if we should call the struct `Text` or `Label`. Again, do we really need to repeat the fact that this is an annotation? `graph.annotations = [TextAnnotation(...), ArrowAnnotation(...)]` just seems a bit longer and more verbose than it needs to be.

  `graph.annotations = [Text(...), Arrow(...), ...]` is much easier to read IMO.

- It would be nice if you could add a test (at least one), so we can all see that it works, and can make sure that nobody accidentally breaks it in the future. Just a simple example showing a chart with a text annotation would be great.

KarthikRIyer:
Stripping the protocol to just the drawing function makes sense.

I think we could call the struct Text, as it could be used for watermarks also and it could be confused with the plot/axes labels. What do you think?
And yeah, simply Text/Arrow is easier to read.

WilliamHYZhang:
That makes sense, I will make the necessary changes today and add a test as well. Thanks!

Sources/SwiftPlot/PlotStyleHelpers.swift
WilliamHYZhang: With @karwa's suggestion of stripping down the Annotation protocol, would this still be applicable? If the Annotation protocol was stripped just to the ```draw``` function, would it make more sense to have ```size``` as a ```Float``` when defining ```Text```?

WilliamHYZhang:
Hi,

Thank you all again for the great feedback on this PR.

First, I've cleaned up and refactored the ```Annotation``` protocol and the ```Text``` struct to reflect the change requests. I agree that it's much cleaner and everything should be working now.

Next, I've added some tests. I created a new folder called ```Annotation``` in the ```SwiftPlotTests``` directory that will eventually contain all the tests for the types of annotations (now it only contains ```annotation-text```). I tried to mimic the style of other test files/folders when writing these tests. In essence it's just an empty line graph with the text watermark on it.

I added corresponding .png and .svg files to agg and svg renderer types, however unfortunately because I don't have ```quartz``` (no MacOs), I couldn't add the reference file for that. Please note that consequently, the test check for ```quartz``` is currently commented out. If someone with access to MacOs and ```quartz``` could run ```swift test``` and add the corresponding outputted file to references that would be much appreciated!

Thanks so much,

William

KarthikRIyer:
@WilliamHYZhang this looks good. But I'd prefer if the example was complete with a plot. It would be nice to know how the annotation would look with a proper graph. Also could you please make the text bigger?
You still need to remove `public struct PlotAnnotation{` from `PlotStyleHelpers.swift` line 23.

Tests/SwiftPlotTests/Annotation/annotation-text.swift
```diff
@@ -30,11 +30,13 @@ extension AnnotationTests {
                                      renderer: agg_renderer)
     verifyImage(name: fileName, renderer: .agg)
     #endif
+    /*
```
KarthikRIyer, line 42: @WilliamHYZhang you need not comment out this part. The #if canImport will take care of it if you're not on macOS.

KarthikRIyer (changes requested):

Sources/SwiftPlot/PlotStyleHelpers.swift
KarthikRIyer: yes

Tests/SwiftPlotTests/Annotation/annotation-text.swift
WilliamHYZhang: The tests seem to fail if this line is not commented out, is that ok?

KarthikRIyer: This test shouldn't fail if it isn't commented out. I tried it. Because of the #if canImport this code shouldn't compile.

WilliamHYZhang: Oh, I was referring to the PR build test. When I had that line uncommented and pushed a commit, I kept on getting an email that the PR test failed.

KarthikRIyer: I'm not sure but that could have been due to the image you had added to the references folder. Could you try uncommenting it?

WilliamHYZhang: I think the error was that the file was not found, if there was no file in the quartz yet the tester still tested it.

KarthikRIyer: Oh alright I understand now. Github actions builds it on macOS. You're right. 
Could you just remove the PlotAnnotation struct, update the example and then someone else can add the reference image for quartz.

odmir: The issue is that the test server does compile the Quartz part and goes and checks if the file is there and it isn’t so it fails. Long term we should come up with a solution for this.

odmir: The issue is that the test server does compile the Quartz part and goes and checks if the file is there and it isn’t so it fails. Long term we should come up with a solution for this.

odmir: I don't understand what happened with the GitHub comments, I commented once, two identical comments were posted for some reason, after refreshing the page they were still there so I deleted one and then the comment of @KarthikRIyer appeared before mine... some caching issue I guess.

WilliamHYZhang: Alright, will do!

WilliamHYZhang:
@KarthikRIyer, I don't see ```public struct PlotAnnotation{``` anywhere in ```PlotStyleHelpers.swift```

WilliamHYZhang:
Ok, pushed update for the tests. Thanks for the review!

KarthikRIyer:
@WilliamHYZhang this looks good. Thanks a lot for the contribution!
Looking forward to see your further contributions in GCI.

Merging.

PR #68, opened 2019-12-18
bfix: Use color alpha while drawing text in SVG Renderer

KarthikRIyer:
We weren't using the alpha of the color while drawing text in svg renderer. I've fixed that here. Noticed it in the watermark example added by @WilliamHYZhang recently.


KarthikRIyer:
@karwa @odmir how is it that `testHistogramStackedStepOffset()` this test runs on macOS without being mentioned in the manifest, but not on Linux. This seems to be the reason test is failing.

odmir:
On macOS it never uses the manifest I think and just uses the Objective-C runtime to discover the tests. The same behaviour should be possible now on Linux with the `--enable-test-discovery` flag though, although only on the latest swift snapshots.

One thing is that there might be a bug in the current implementation which we would run into:
https://bugs.swift.org/browse/SR-11951

karwa:
Yup, macOS uses the Obj-C runtime to discover tests. It’s annoying, but test discovery apparently isn’t on by default on Linux (yet)

KarthikRIyer:
I'm running into the bug @odmir mentioned above. If I merge this PR could either @karwa or @odmir add the test image in another PR?

odmir:
Sure, go ahead!

KarthikRIyer:
@odmir done.

PR #71, opened 2019-12-22
Add Arrow Annotation

WilliamHYZhang:
Hi,

I've added a simple arrow structure to annotations. In essence, it consists of using the ```drawLine``` function from the renderer once for the arrow body and 2 times for the arrow head.

The arrow is quite flexible, it allows for specifying the color, start and end points, width, length of the arrow head, and angle of the arrow head relative to the arrow body.

Having ```drawLine``` draw the body of the arrow is quite trivial, as the start and end points, etc. are already provided. However, drawing the arrow head is a bit more tricky. We already know the start point of each part of the arrow head will be start point of the arrow itself, however in order to know the end point of each part of the arrow head, we use some mathematics.

For the sake of simplicity, we assume that the arrow is level to the x axis pointing to the left. Then, the two endpoints of the arrow head can be calculated as:
```
p1 = start + Point(cos(headAngle)*headLength, sin(headAngle)*headLength)
```
and
```
p2 = start + Point(cos(headAngle)*headLength, -sin(headAngle)*headLength)
```

Then, we can rotate these two points by the angle between the arrow body and the x-axis to get the location of where the end points should be. To find this angle, we use the ```atan2``` function as follows:
```
rotateAngle = -atan2(end.x - start.x, end.y - start.y)
```

Afterwards, we can use the helpful ```rotatePoint``` function to rotate these two points by the angle we just found and then call the ```drawLine``` function to draw the arrow head.

Here's the final result:
image

I specified different colors, angles, sizes, etc. and it seems to work great. Hope you guys like it!

William


KarthikRIyer:
@WilliamHYZhang a few things I want to point out:

- I think it's a little more intuitive to have the arrow at the end point instead of start point.
- I would increase the default angle to 20 and increase the default width to something more comfortable to the eyes. I'm also thinking it might be better to call it strokeWidth instead.
- You also need to plot the arrows in the co-ordinate space of the plot, not the absolute coordinates of the image. Take a look at how the plot lines are plotted in LineChart.
- I  would update the example/test to plot a simple sin curve and instead of many arrows, we could have just one arrow pointing to a local maxima and a text annotation saying local maxima. Something like the example image [here](https://matplotlib.org/3.1.1/tutorials/text/annotations.html#basic-annotation).

WilliamHYZhang:
Thanks for the suggestion! Will implement soon.

odmir:
Very interesting!
Maybe I can suggest some (cool) extra features you can implement if you think any of them is interesting, and if it is ok with you and @KarthikRIyer:

- Different kinds of arrow head. Have an option to specify between filled arrow head or just the "skeleton" like you have now.
- Allow for an optional `Annotation` at the start (and maybe also the end) of the arrow. You would then in the `draw` method of the arrow call the `draw` method of the optional `Annotation` at the coordinate of the start (or end) of the arrow. This would be cool because then you can compose for example a `Text` annotation with the arrow, or eventually have a `Box` annotation with a `Text` annotation inside, in one of the tips of the `Arrow`. Also maybe as another future addition, add the ability to choose which side of the optional annotation is going to be glued to the start (or end) of the arrow (where it is going to be anchored) i.e. `.east`, `.west`, `.north`, `.south`.

These are just a couple ideas I had, you do it only if you’d like to and you would be free to do these in a new PR. This if @KarthikRIyer agrees, of course, and we would probably need to give you more pointers as well.

odmir:
Maybe in the future it would be interesting to have all annotations have optional anchors where you can attach other annotations and specific annotations could have more specific anchors (like the start and end of an `Arrow` or the leading/trailing edge/baseline of a `Text` annotation.
This is an idea for the future though, not for now 😅

KarthikRIyer:
These ideas are very interesting! I'd definitely like to see them implemented. It would be nice if you could implement the filled arrow head feature for GCI but the rest is up to you @WilliamHYZhang. Maybe you could leave out implementing box/rectangle annotation and instead implement the arrow features @odmir suggested. So in the end we could have a more feature complete annotation type. You're free to implement what you like.

WilliamHYZhang:
Thank you for the great ideas @odmir! My flight just landed in China, but once I settle in I'll get started working on this.

WilliamHYZhang:
Alright, I'm very jetlagged at the moment so I just pushed some minor updates that addressed some of the requested changes, now the start and end points should be updated to reflect @KarthikRIyer's comment. Also updated the default angle and variable names, as well as the test code.

I'm confused by what you mean regarding the third bullet point of "You also need to plot the arrows in the co-ordinate space of the plot, not the absolute coordinates of the image. Take a look at how the plot lines are plotted in LineChart.", would I be correct in assuming that I need to scale the points similar to lines 404 to 415 in ```LineChart.swift```, or am I completely misunderstanding this?

Also, I can't seem to figure out why, but changing the ```strokeWidth``` variable doesn't have any effect on the arrow. I'm calling ```drawLine``` and passing ```strokeWidth``` as an argument so I'm a bit concered why it doesn't show any changes.

PR test is failing since I didn't update the image in the references folder, no need to worry about that. Will update tomorrow once I get these more pressing issues sorted out.

Hopefully tomorrow I will be well-rested enough to thoroughly read @odmir's comments about more arrow features and implement them, thanks again for the great suggestions.

William

Sources/SwiftPlot/PlotStyleHelpers.swift
```diff
@@ -67,3 +67,51 @@ struct Text : Annotation {
...
+    public var headLength: Float = 10
+    public var headAngle: Float = 20
+    public func draw(renderer: Renderer) {
+        // Draws arrow body.
+
+        //TODO: scale points to be plotted according to plot size
+        
+        renderer.drawLine(startPoint: start,
+                          endPoint: end,
+                          strokeWidth: strokeWidth,
```
WilliamHYZhang: Changing "strokeWidth" seems to have no effect on the output.

Tests/SwiftPlotTests/Annotation/annotation-arrow.swift
```diff
@@ -11,46 +12,32 @@ import QuartzRenderer
...
-    let x:[Float] = [10,100,263,489]
-    let y:[Float] = [10,120,500,800]
+    func function(_ x: Float) -> Float {
+      return sin(x)
+    }
 
     let lineGraph = LineGraph<Float, Float>(enablePrimaryAxisGrid: true)
-    lineGraph.addSeries(x, y, label: "Plot 1", color: .lightBlue)
-    lineGraph.plotTitle.title = "SINGLE SERIES"
+    lineGraph.addFunction(function, minX: -5.0, maxX: 5.0, label: "sin(x)", color: .lightBlue)
```
odmir: Here instead of wrapping `sin` in `function` and using that, you can use `sin` directly as the first parameter since it has the same type signature.

Sources/SwiftPlot/PlotStyleHelpers.swift
odmir: You have just stumbled into a bug in CPPAGGRenderer! If you take a look at the output for SVG you can see your changes are taking effect. If you want you can fix it by going to CPPAGGRenderer.cpp and write `stroke.width(thickness);` in line 350.

Sources/SwiftPlot/PlotStyleHelpers.swift
```diff
@@ -67,3 +67,51 @@ struct Text : Annotation {
...
+        renderer.drawLine(startPoint: end,
+                          endPoint: p1,
+                          strokeWidth: strokeWidth,
+                          strokeColor: color,
+                          isDashed: false)
+        renderer.drawLine(startPoint: end,
+                          endPoint: p2,
+                          strokeWidth: strokeWidth,
+                          strokeColor: color,
+                          isDashed: false)
```
odmir: Here you can use `renderer.drawPlotLines` and pass it an array of the points so that for higher thicknesses it doesn't get like this:
Screenshot 2019-12-23 at 12 47 46

Sources/SwiftPlot/PlotStyleHelpers.swift
WilliamHYZhang: Thank you for noticing this! I'll create an issue for this and a PR to fix it.

WilliamHYZhang:
Ok, pushed with update.

- Changed some default values to make arrow look better
- Used ```drawPlotLines``` instead of the buggy ```drawLine``` as per @odmir's suggestion, also should resolve @KarthikRIyer's comment that "You also need to plot the arrows in the co-ordinate space of the plot, not the absolute coordinates of the image. Take a look at how the plot lines are plotted in LineChart". Fixes the ```strokeWidth``` issue as well.
- Added option for dashed arrow
- Added option for filled arrow head, implemented with ```drawSolidPolygon```
- Updated test case with new graph with sin function + arrow annotation pointing to top of sin function + text saying "relative maxima"

Sources/SwiftPlot/PlotStyleHelpers.swift
WilliamHYZhang: Update: see #73

KarthikRIyer:
This looks good @WilliamHYZhang. Do you plan to implement other features that @odmir suggested, the rectangle/box annotation or anything else? What's your plan of action?
If you plan to complete the arrow annotation implementation I'd prefer it to be in this PR itself. If you plan to work on anything else, I think this should be ready to mere.

WilliamHYZhang:
I plan on implementing the text anchor addition to the arrow structure for this PR today, as well as a simple rectangle/box support for text (that will be a new PR). Once the text anchor is done this PR should be ready to merge.

Thank you!

WilliamHYZhang:
Strange, the tests seem to be all passing on my local machine, running ```git status``` shows that I'm up to date as well.

WilliamHYZhang:
Alright, everything should be working now.

I added the option for text annotation and N/E/S/W anchor point specifications. I used the text properties such as the width, height, etc. of the text to obtain the correct offsets to pass for the location of the text when calling ```renderer.drawText()```. 

I also updated the test for the arrow annotation to use this new feature.

William

KarthikRIyer:
@WilliamHYZhang what @odmir meant by having more annotation wasn't to add a text explicitly to arrow annotation but to have a generic annotation type in it so that you could have either a text, box or maybe another annotation anchored to the start or end of any other annotation.

Although I'm not sure what @odmir meant by north south east west. Could you clarify that please?

So, William what you would need to do is in the arrow annotation have a start Annotation and an end Annotation, which can be optionals. You can set the location of these optionals inside the arrow annotation itself and just call their draw method after drawing the arrow.

odmir:
> Although I'm not sure what @odmir meant by north south east west. Could you clarify that please?

Ok, this was meant as a way to specify which side of the annotation would be drawn at the starting (or ending) coordinate of the arrow. If you specify `.north` for example, then the annotation would be drawn so that the arrow is coming out from the top. But I think this should be designed a little more, before implementing it in `Arrow`.

@WilliamHYZhang, what @KarthikRIyer said about what I meant is correct. Instead of hardcoding that the arrow can only work with a `Text` annotation, it can be written in a way that allows any annotation to be drawn at either end of the arrow.
Remember what I said about every `Annotation` knowing how to draw themselves? Well that comes in handy now as well because if you want this feature to work with any annotation you can ask whatever annotation was specified to draw itself at the start or end of the arrow.

I would say that for right now you do not need to worry about the N/E/S/W anchors as it should be discussed further how this should work in a way that works for all annotations, current and future. This will lead to the annotations drawn at either end of the arrow to overlap with the arrow itself in some cases, but maybe for right now that is ok, just until the details around this are a little bit more fleshed out.

odmir:
Here is a link to help you grasping a little better what `Optional`s are and how they work in swift:
https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html

Go down to the "Optionals" section and read just until you reach "Error Handling". This is the basics on optionals and it is actually almost all there is to know about them.

Then there is this:
https://docs.swift.org/swift-book/LanguageGuide/OptionalChaining.html

Which you can read once you've grasped `Optional`s a little better and practiced maybe by using them to implement the optional `Annotation`s on `Arrow`. You don't need to know everything that is in there, but some things will certainly come in useful for calling the `draw` method on the optional annotations.

WilliamHYZhang:
@KarthikRIyer @odmir thank you all for the comments and guidance. I will read up on optionals and how to implement them today.

> You can set the location of these optionals inside the arrow annotation itself and just call their draw method after drawing the arrow.

With what @odmir said about N/E/S/W anchors, in the future, would it be ideal to have the annotations draw themselves based on the anchor points? For example, I already have the code to do that for the ```Text``` annotation, and that could be easily modified to generalize for any ```Annotation``` with an obtainable width and height.

Thanks again,

William

WilliamHYZhang:
Alright, pushed with the new update, now drawing annotations on either start/end for any annotation type is supported with ```Optional```s.

odmir:
> would it be ideal to have the annotations draw themselves based on the anchor points?

Yes, in case we are going forward with using anchors in the future, the annotations would draw themselves based on the anchors. Right now you have to specify the position of the `Text` annotation so that it draws at either end of the arrow, but ideally you should not need to tell them the position manually, they themselves should be able to figure that out, but for that to work the arrow needs to "tell" the other annotation where its start or end point are so that they draw at the right place. This is one of the things that should be discussed further, right now to do that you would have to add an origin and a size to the `Annotation` protocol to be able to do it, but I don't think that is the right way.

So ideally, in the future, we would just say that we want an `Arrow` with a `Text` annotation attached to the starting anchor of the arrow by the right anchor of the text for example, and the `Arrow` annotation should not need to know anything about the size or position of the `Text` annotation (or any other annotation).

odmir:
Btw, good job on the optionals! For right now I think this suffices.

> You also need to plot the arrows in the co-ordinate space of the plot, not the absolute coordinates of the image. Take a look at how the plot lines are plotted in LineChart.

@WilliamHYZhang what I think @KarthikRIyer means by this is that right now you are specifying the coordinates for the position of the annotations in the coordinate space of the image itself. The images in the tests have dimensions of 1000x600 px and because you are using the coordinate space of the image when you draw the `Arrow`, you have to specify the start and end points of the arrow like this: Point(400.0, 585.0) and Point(585.0, 585.0). If you instead use the coordinate space of the plot (the space of the represented data itself) you could have specified the arrow coordinates as follows: Point(-.pi/2, 1.0) and Point(.pi/2, 1.0) which is a more natural way to use arrows when you want to point them at a specific point of the represented data itself, in this case the maximum of the sine function (x = pi/2; y = 1).

That said I think there's value in being able to draw arrows (and any annotation for that matter) in other coordinate spaces as well, like the one of the image itself and also a normalised space where every point on the final plot is between 0 and 1. This could be specified as an option somewhere in the future, when this is discussed further.

WilliamHYZhang:
Thanks for the great insight and suggestions! I hope they will be implemented soon.

This is just an idea for the "drawing by itself" for the arrow annotations, what if we have a function in each annotation that when passed the arrow, side, and anchor would do the necessary calculations and draw itself there? Just some food for thought.

KarthikRIyer:
Yes, @odmir is right. I seem to have missed to clarify what I meant when I talked about the coordinate space.
You need to scale the points according to the plot's coordinate space like at line 339 in LineChart.swift.
You can have an enum like in HistogramSeriesOptions.swift with options like imageSpace, plotSpace and ndc (Normalized Device Coordinates) and you can scale accordingly. This will probably require some more thought. You would need to think about the default arrow sizes when using each coordinate system and also how you would get the plot dimensions to the annotation. Also you might need to move the drawAnnotation function out of the drawForeground function to here:
```
renderer.withAdditionalOffset(results.plotBorderRect.origin) { renderer in
            drawData(markers: results.plotMarkers, size: results.plotBorderRect.size, renderer: renderer)
        }
```
when drawing in the plot's coordinate space or NDC, because you need to add the x and y offset to each point for origin shifting.

WilliamHYZhang:
> That said I think there's value in being able to draw arrows (and any annotation for that matter) in other coordinate spaces as well, like the one of the image itself and also a normalised space where every point on the final plot is between 0 and 1. This could be specified as an option somewhere in the future, when this is discussed further.

Yeah, I can understand the option of that. I'll see if I can make a new PR afterwards that reworks the current annotation types to support this.

EDIT: I'll just push it onto this PR and update the text annotation in a later PR.

WilliamHYZhang:
> 
> 
> Yes, you @odmir is right. I seem to have missed to clarify what I meant when I talked about the coordinate space.
> You need to scale the points according to the plot's coordinate space like at line 339 in LineChart.swift.
> You can have an enum like in HistogramSeriesOptions.swift with options like imageSpace, plotSpace and ndc (Normalized Device Coordinates) and you can scale accordingly. This will probably require some more thought. You would need to think about the default arrow sizes when using each coordinate system and also how you would get the plot dimensions to the annotation. Also you might need to move the drawAnnotation function out of the drawForeground function to here:
> 
> ```
> renderer.withAdditionalOffset(results.plotBorderRect.origin) { renderer in
>             drawData(markers: results.plotMarkers, size: results.plotBorderRect.size, renderer: renderer)
>         }
> ```
> 
> when drawing in the plot's coordinate space or NDC, because you need to add the x and y offset to each point for origin shifting.

Thank you @KarthikRIyer and @odmir for the helpful clarifications. If I'm understanding this correctly, will I have to include another boolean argument ```isOriginShifted``` to and handle the cases to draw with/without shifted origin?

My main concern would be where to put this code. Would it be better to have a function that simply converts coordinate points to absolute points on the image, and use that function for all the annotations that will eventually use them? If so, would I put that in ```PlotStyleHelpers.swift``` or another file?

Thanks,

William

WilliamHYZhang:
Also, would it be better to generalize this "coordinate points to absolute point" function so that even files such as ```LineChart.swift``` can utilize them?

KarthikRIyer:
Yes I suppose you could generalize this conversion. You would always need the output to be in image space so you could have a function that takes in a point, the plot dimensions the input coordinate system and returns the point in the specified output coordinate system. 
If this function is to be used by the plots too then it would be better to put it under HelperFunctions.

It would be better to discuss this with others before moving forward. @odmir @karwa?

WilliamHYZhang:
Ok, sounds good. Since this function would be applied to all annotations and some other files as well, would it be better to first merge the two annotation PR's once they are good to go, and then I can open up a new PR for this conversion as another task? I just don't want to clog up the two PR's and deal with merge conflicts or any other issues right away. Afterwards, I can work on implementing anchor support and other more advanced features for the annotations.

KarthikRIyer:
Alright. I'm merging this. You can pull in master and resolve any conflicts for your Box PR on your system itself.

WilliamHYZhang:
See #76 for the issue regarding coordinate conversion, please continue the conversation there.

PR #73, opened 2019-12-23
bfix: fix draw_line bug

### LINKED ISSUE #72: strokeWidth argument when calling renderer.drawLine() not working with AGGRenderer
Discovered this while working on the arrow annotations, changing the width of the line seems to have no effect on the line width in the output with AGGRenderer.

WilliamHYZhang:
Should resolve #72. Thanks to @odmir for suggesting this change.


WilliamHYZhang:
LOL, seem like we have 29 failures for the PR test build now. I believe that this is because the previous images in annotations are now incorrect, for example looking at some test cases the lines are now super thicc. Can we replace these images and update the test cases?

WilliamHYZhang:
If someone with access to MacOs and ```quartz``` renderer (along with the other renderer types) could fork my ```bfix``` branch, run ```swift test```, and overwrite the ```Reference``` folder with the produced ```output``` folder that would be much appreciated!

EDIT: Oh wait, this bug is with the AGGRenderer only. Whoops lol.

Sources/AGGRenderer/CPPAGGRenderer/CPPAGGRenderer.cpp
```diff
@@ -347,6 +347,7 @@ namespace CPPAGGRenderer{
       agg::conv_transform<agg::path_storage, agg::trans_affine> trans(rect_path, matrix);
       agg::conv_curve<agg::conv_transform<agg::path_storage, agg::trans_affine>> curve(trans);
       agg::conv_stroke<agg::conv_curve<agg::conv_transform<agg::path_storage, agg::trans_affine>>> stroke(curve);
+      stroke.width(thickness);
```
WilliamHYZhang, line 350: What I believe is happening here is that all of the previous test cases that have been tested don't take into account the stroke width when using the draw_line function, so when I add this line, virtually all of the AGGRenderer test cases take the L.

WilliamHYZhang:
Ok, pushed with overwrite to the ```Reference/agg``` folder, as well as encoding test check for AGGRenderer.

KarthikRIyer:
Looks good. Thanks! Appreciated.

PR #84, opened 2020-01-03
feat: update notebook

WilliamHYZhang:
This updates the notebook with @BradLarson's suggested changes on the S4TF Google Group. Currently, there's a bug in LineChart that prevents scaling if the original loss isn't plotted along with some other wacky things #81, so I just added a new line, "Scaled Test", that is the test loss scaled to the train loss, so they can be better compared.


Notebooks/Machine Learning with Swift for TensorFlow and SwiftPlot.ipynb
```diff
@@ -1167,6 +1189,8 @@
     "\n",
     "lossGraph.addSeries(epochs, trainLoss, label: \"Train\", color: .orange)\n",
     "lossGraph.addSeries(epochs, testLoss, label: \"Test\", color: .lightBlue)\n",
+    "let scaledTestLoss = testLoss.map {$0 * 6}\n",
```
BradLarson: Rather than applying a map to perform a sixfold scaling here and introduce a third series, I'd highly recommend applying loss averaging at the point above where losses are added to the array.

BradLarson:
My recommendation as a robust way to solve this would be to add a `count` property to Stat, increment that every time you do a += on the loss property, and then instead of appending `train|testStat.loss`, append `train|testStat.loss / Float(train|testStat.count)`. That will give you the true average loss values (which should be fairly small numbers) and won't require the extraneous series to be plotted.

You can then remove any scaling at plotting time, because they both should be in the same scale at that point.

WilliamHYZhang:
Yeah, that was my original idea as well, however the small numbers are resulting in absolutely nothing showing up in the chart.

WilliamHYZhang:
Because the batch size I chose for the tutorial divides evenly for both splits, my original implementation for your suggestion was to append `train|testStat.loss / Float(120|20)`. Those values should always be the number of batches for each epoch respectively for each split.

BradLarson:
Dividing by a constant number for the test / train batches when appending loss should work as well as accumulating a count. The count is slightly more flexible with varying batch size, but you shouldn't have a problem hardcoding that.

The plot should be able to show the loss values correctly, because we've plotted values in this range before using the framework. Is there something off in the logic for setting the default Y range? Can you manually specify a 0.0 - 2.0 (or whatever the loss max is) for a Y range?

WilliamHYZhang:
Yes, that's what I'm suspecting. I'll look into what I can come up with. There is a similar argument for what you are describing, `clampY` (e.g. `clampY = -50...50`, however that is for `addFunction` and not `addSeries`. I'll see if I can implement something like that for `addSeries` and if it fixes the issue.

WilliamHYZhang:
@KarthikRIyer, #86 should fix these issues, please merge that and I'll rewrite it the tutorial with the updated codebase with the correct losses. Thanks!

KarthikRIyer:
Done @WilliamHYZhang

WilliamHYZhang:
Thanks @KarthikRIyer for merging and @karwa for fixing the LineChart errors, now the graphs look very nice. Pushed with update.

image

KarthikRIyer:
@WilliamHYZhang looks good.

PR #86, opened 2020-01-03
LineGraph fixes

### LINKED ISSUE #81: Missing points, strange errors in line graph

karwa:
Fixes #81 

The layout data for each axis is now calculated separately by a helper struct, then we merge their X-axes (we could also _not_ merge those axes, and show the second X-axis on top, but we can add that later).

The reference images have changed because the origin is now calculated as (10, 10) for some plots (the first data point), rather than (0,0). It would be nice to snap this to a nicer value again, but that's a bit complex to do right. In the mean time, this fixes a bunch of other issues and adds new tests for a bunch of weird cases.


karwa:
Oh I have a slight improvement to make to this...

karwa:
OK, we now snap the origin to 0 if it happens to be in-view after calculating the axis scale based on the data (i.e. if it's in the margins or something). Changes to tested cases are barely perceptible, except for `_07_secondary_axis_line_chart`; since 0 on the Y-axis is not in view for the primary series (which starts at (10, 150)), the primary series Y markers are different.

KarthikRIyer:
Merging

PR #87, opened 2020-01-04
Update docs for Google Colab support

WilliamHYZhang:
Excited to announce that I've found a way to use SwiftPlot with inline graph display support for Google Colab. Although this wasn't a GCI task I think it would be really helpful and count for the community involvment aspect of GCI, perhaps maybe I could continue to work on this feature in a future tasks.

Anyways, you can see my Google Colab environment [here](https://colab.research.google.com/drive/13iMA72WQyFwMiarrCJqfxkxCTxxUxsiq).

Obviously this is not perfect yet, but it's definitely a start! Hope you all enjoy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ SwiftPlot currently uses three rendering backends to generate plots:
 
 To encode the plots as PNG images it uses the [lodepng](https://github.com/lvandeve/lodepng) library.
 </br>
-SwiftPlot can also be used in Jupyter Notebooks.
+SwiftPlot can also be used in Jupyter Notebooks, with ongoing development for Google Colab.
 </br>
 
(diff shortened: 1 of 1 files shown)


KarthikRIyer:
Great effort @WilliamHYZhang! This is a good start. But actually if you want to rely on Python, there's a better way to do it. You've included the EnableIPythonDisplay file. I had added a function to display base64Png images during my GSoC and when I got the pure Swift version I removed the Python function.

https://drive.google.com/open?id=1kjbLoqcwD3BDF0i2LSRg9GW5z902Rl5c

See the first image that is displayed and the `display` function. This function was earlier in the EnableIPythonDisplay file. The blurriness issue is also solved here.

WilliamHYZhang:
Cool! Thanks so much for sharing, I'll update the README.md to reflect
this.

KarthikRIyer:
Thanks for the PR! Appreciate it!

PR #89, opened 2020-01-07
remove installation output

WilliamHYZhang:
per suggestions from the S4TF Google Group


WilliamHYZhang:
Odd, there seem's to be a syntax error in the ipynb file that I can't seem to find.

PR #91, opened 2020-01-10
Remove [-Wswitch] warning during import

boronhub:
A small commit, but my first to this project nevertheless !

Removed redundant `end_of_markers` case declaration in `AGG/include/agg_renderer_markers.h`, which gets rid of the following warning during import :

```
warning: enumeration value 'end_of_markers' not handled in switch [-Wswitch]
            switch(type)
                   ^
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@
       * [Line Graph with Sub Plots stacked horizontally](#line-graph-with-sub-plots-stacked-horizontally)
       * [Plot functions using LineGraph](#plot-functions-using-linegraph)
       * [Using a secondary axis in LineGraph](#using-a-secondary-axis-in-linegraph)
+    * [Bar Graphs](#bar-graphs)
+      * [Simple Bar Graph](#simple-bar-graph)
+      * [Stacked Bar Graph](#stacked-bar-graph)
     * [Displaying plots in Jupyter Notebook](#displaying-plots-in-jupyter-notebook)
   * [How does this work ?](#how-does-this-work)
   * [Documentation](#documentation)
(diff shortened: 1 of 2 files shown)


WilliamHYZhang (commented):
Great to see other GCI people contributing to SwiftPlot! @bornhob, SwiftPlot uses AGG (see [here](https://github.com/tomhughes/agg)) as one of its renderers. The file you are editing is directly from that repository, so there is a possibility that `end_of_markers` was included there for a specific reason. However, build test seems to pass, so let's see what @KarthikRIyer has to say.

KarthikRIyer:
@boronhub the code definitely compiles on removing end_of_markers. From what I can make out it is there to make the code more readable. I would prefer not to make too many changes unless its very necessary. Just having a `default:` case in switch should suppress the warning. I wouldn't remove anything from the enum.

karwa:
The better solution would be to use an updated version of AGG. We're only using 2.4, which produces a lot of warnings that appear to be silenced by 2.6 (e.g. this issue is fixed [here](https://github.com/ghaerr/agg-2.6/blob/master/agg-src/include/agg_renderer_markers.h#L575) by adding the `end_of_markers` case).

FWIW, it doesn't seem that AGG will ever get any releases beyond 2.6, for [rather tragic reasons](https://vector-agg-general.narkive.com/1wMI33Jj/r-i-p-maxim-shemanarev). We might as well update to the _final_ version of AGG.

KarthikRIyer:
I used 2.4 because matplotlib uses that version and also I was having some trouble building with other versions that time.
2.6 doesn't seem to be an official release but it does seem to fix the warnings. When I tried building I got no warnings related to AGG.

#93 
Although the tests seem to fail with the update.

boronhub:
> 
> 
> @boronhub the code definitely compiles on removing end_of_markers. From what I can make out it is there to make the code more readable. I would prefer not to make too many changes unless its very necessary. Just having a `default:` case in switch should suppress the warning. I wouldn't remove anything from the enum.

`default ` isn't suppressing the warning on it's own, I tried.

PR #92, opened 2020-01-10
Update README.md with BarGraph usage examples

boronhub:
Also added structure to include usage examples for all graph types

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@
       * [Line Graph with Sub Plots stacked horizontally](#line-graph-with-sub-plots-stacked-horizontally)
       * [Plot functions using LineGraph](#plot-functions-using-linegraph)
       * [Using a secondary axis in LineGraph](#using-a-secondary-axis-in-linegraph)
+    * [Bar Graphs](#bar-graphs)
+      * [Simple Bar Graph](#simple-bar-graph)
+      * [Stacked Bar Graph](#stacked-bar-graph)
     * [Displaying plots in Jupyter Notebook](#displaying-plots-in-jupyter-notebook)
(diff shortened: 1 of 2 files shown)


WilliamHYZhang:
@boronhub do you mind making a separate branch in your fork for your two pulls? Thanks!

In order not to clog up the README, maybe instead we can just direct users to the testing folder where they can browse code examples and output of all of the graphs?

KarthikRIyer:
I agree with @WilliamHYZhang. You should have a separate branch for each PR that you work on.

Also, because we have a dedicated Tests/examples directory included with the repository we shouldn't need to have examples for everything in the README. Instead of having these extra examples, you could just add something like this:

For examples of other plot types and features refer to the [Tests](https://github.com/KarthikRIyer/swiftplot/tree/master/Tests/SwiftPlotTests) directory.

boronhub:
Apologies for the branch thing, I created a new one but I guess I forgot to push that one.

And yes, I also had the feeling the README would be too dense, so I wanted to know what @KarthikRIyer thought of maybe reducing the number of LineGraph examples, and have maybe one/two usage examples per graph type (most common uses) .

WilliamHYZhang:
+1 on @KarthikRIyer's suggestion.

KarthikRIyer:
@boronhub I'd like to keep the current examples because they demonstrate some of the basic features of the framework, the usage of which will be similar in other plots. Take sub plots for example. The method to create sub plots will remain the same everywhere. If you want to add more series to bar graph the syntax remains almost same.

KarthikRIyer:
@boronhub would you like to add the link to the tests directory to the README?

Issue #75, opened 2019-12-28
Create single renderer function to draw Rect with fillColor and border


KarthikRIyer:
> Our renderers right now can only draw a filled rectangle separate from a stroked one which is not ideal. Every renderer out there can draw a rectangle with both a fill color and a stroke color at once, instead of having to draw two rectangles one on top of the other with different settings. We should probably replace drawRect and drawSolidRect by a simpler one that accepts a strokeWidth, strokeColor, fillColor and hatchPattern.

boronhub:
Can I work on this ?

KarthikRIyer:
Sure @boronhub !

boronhub:
```
public func drawSolidRectWithBorder(_ rect: Rect,
                                        strokeWidth thickness: Float,
                                        fillColor: Color = Color.white,
                                        borderColor: Color = Color.black) {
        var x = [Float]()
        var y = [Float]()

        let pts = getPoints(from: rect)
        x.append(pts.tL.x + xOffset)
        x.append(pts.tR.x + xOffset)
        x.append(pts.bR.x + xOffset)
        x.append(pts.bL.x + xOffset)
        y.append(pts.tL.y + yOffset)
        y.append(pts.tR.y + yOffset)
        y.append(pts.bR.y + yOffset)
        y.append(pts.bL.y + yOffset)

        draw_solid_rect(x,
                        y,
                        fillColor.r,
                        fillColor.g,
                        fillColor.b,
                        fillColor.a,
                        0,
                        agg_object)
        draw_rect(x,
                  y,
                  thickness,
                  borderColor.r,
                  borderColor.g,
                  borderColor.b,
                  borderColor.a,
                  agg_object)
    }
```


isn't this what is required from this task ?

KarthikRIyer:
Yes @boronhub . But you need to add the function to the Renderer protocol first and then add implementations to each Renderer.

boronhub:
And in my PR, do I have to remove the original two rectangle functions or no?

KarthikRIyer:
Don't remove those functions. They might come in useful somewhere.

boronhub:
Dosen't `drawSolidRectWithBorder` fit this criteria ? It's working right now across all three renderers. It also exists in Renderer.swift

WilliamHYZhang:
[...]

PR #95, opened 2020-01-15
Added single renderer function to draw Rectangles

boronhub:
PR for #75 as part of GCI.

Added `draw_solid_rect_with_border ` to AGG renderer so that calling `drawSolidRectWithBorder`does not require overlaying two rectangles made with `draw_rect `and `draw_solid_rect `.


Sources/AGGRenderer/AGGRenderer/AGGRenderer.swift
```diff
@@ -95,22 +95,19 @@ public class AGGRenderer: Renderer{
...
                         y,
+                        thickness,
+                        fillColor.r,
+                        fillColor.g,
+                        fillColor.b,
+                        fillColor.a,
                         fillColor.r,
                         fillColor.g,
                         fillColor.b,
                         fillColor.a,
```
KarthikRIyer: Why are you using fillColor here? Shouldn't it be borderColor?

KarthikRIyer (changes requested):

KarthikRIyer:
@boronhub are you on linux or mac?

boronhub:
linux

boronhub:
@KarthikRIyer how can i fix the build errors?

WilliamHYZhang:
@boronhub looking at the error: `ld: symbol(s) not found for architecture x86_64`, the build is probably experiencing an error compiling some code. I don't know about the specifics, though. is `swift build` and `swift test` passing on your local system?

EDIT: probably something to do with the recent merge to update the AGG version... see #93, tests are failing after that push.

boronhub:
Yeah, it's passing.

Sources/AGGRenderer/CPPAGGRenderer/CPPAGGRenderer.cpp
```diff
@@ -279,6 +279,35 @@ namespace CPPAGGRenderer{
       agg::render_scanlines(m_ras, m_sl_p8, ren_aa);
     }
 
+    void draw_solid_rect_with_border(const float *x, const float *y, float thickness, float r_fill, float g_fill, float b_fill, float a_fill, float r_stroke, float g_stroke, float b_stroke, float a_stroke){
```
karwa, line 282: You're missing a `const void *object` parameter here, which is why the build is failing. Your header declared that such a function exists, but the linker couldn't find it because this parameter is missing.

boronhub:
@KarthikRIyer @karwa the tests are failing because of an `XCTAssertTrue failed - 🔥 Image mismatch`, does that have to do with the recent AGG update ? Even one of my other PRs #96 is failing, even though it is just the addition of one line to the README.

karwa:
No, the tests are failing here because the output has changed. The thickness of the legend border seems different.

image

boronhub:
@karwa what's triggering this?

karwa:
hard to say - you'll have to debug it. You might try doing the stroke _after_ the fill, though.

WilliamHYZhang:
@boronhub once @karwa's PR gets merged try updating your fork against upstream after you fix the local errors.

karwa:
That PR won't fix the issue - it only addresses a bug where some tests were not being run on Linux. These are different tests that are failing, and they are failing for good reason: the output has demonstrably regressed.

The legend's default fill colour is not transparent - it is actually white with 70% alpha. If you draw the border first, then fill the same rectangle on top of it, you're going to be drawing _over_ the inner half of the border. Instead, you should fill the rectangle and draw the border on top of it.

boronhub:
@karwa I made that change, but the tests are still failing.

boronhub:
@karwa ping

KarthikRIyer:
@boronhub sorry for the delay. I've been occupied elsewhere and will probably be occupied this weekend. I've extended your deadline by 2 days just in case I'm unable to get to this.

KarthikRIyer:
I just pulled your changes on my system and ran swift test. All tests seem to pass except the line chart crossBothAxes and it seems to be because the X range of the plot drawn is 0 to 5 and the reference is -5 to 5.

boronhub:
Thanks @KarthikRIyer! How do I fix the inconsistencies?

boronhub:
@KarthikRIyer @karwa merging the latest commits made the tests pass !

KarthikRIyer:
Look good @boronhub. Thanks for your work and patience. Appreciated!

PR #97, opened 2020-01-16
Test fixes for recent AGG update

karwa:
- Updated linuxmain
- Updated reference images (for images which changed with recent AGG update, but were not regenerated due to missing entries in linuxmain)
- Use `renderAndVerify` helper in histogram tests (some tests were not being verified).

In future I think we should probably prefer to use `renderAndVerify` for the tests - it ensures we don't make any mistakes (like the histogram tests, forgetting to verify), makes the tests easier to read, and makes it easier to add/remove renderers in the future if we want to.

It might also be good to just move all the "example"-style tests out to their own file, where they can show the rendering steps if they want. As our tests become more complex, it's less clear that it's a good idea to mix complex tests (like `histogram_multi_stacked`) with simpler, example-style tests.


KarthikRIyer:
Thanks a lot @karwa. I like the renderAndVerify method. It does make the work easier!

PR #99, opened 2020-01-18
Add a list of built-in colors

areebg9:
I feel like it would be helpful for the list of available colors to be readily available in the README, under the Colors section. It may be better to instead of listing them all in a table, just write it out in text, e.g.

translucent, black, gray, ...

Let me know which one you think is the better option (the current PR just adds a table - the only issue is that it takes up lots of vertical space).


WilliamHYZhang:
Yeah, this does take up quite a bit of space. README already needs quite a bit of scrolling as is, maybe just adding a link to where the color profiles are defined would be better.

KarthikRIyer:
I agree with @WilliamHYZhang. The list of colors is a very trivial thing that any developer can lookup in the library's code. In case we really want top document it it would be better to just add a link to the file where the colors are mentioned. If we want a full list then it should be in a wiki or some other kind of detailed docs. The frameworks isn't very big/complex at the moment so having such a detailed wiki doesn't make sense.

areebg9:
Alright. I was just thinking that users shouldn't have to go into the code to get such simple information, so I was thinking we could alternatively just list them out (like a list) and not in a chart. If we don't need this at the moment, feel free to close this PR.

KarthikRIyer:
Why don't you add a link to the file where the colors are defined? I think that would be helpful too.

areebg9:
Sure

areebg9:
Alright, I've changed it. Now it just says "Built-in colors can be found here," with a link to Color.swift.

KarthikRIyer:
Thanks @Qwerty71 !

PR #100, opened 2020-01-19
Add Anchor Support for Annotations

WilliamHYZhang:
Introducing the newest feature for annotations, the ability to anchor them to anything! This idea started as just to anchor text annotations, but with the help of some nifty tricks we can expand this significantly. Here's an overview of what I did in this PR:

- Implement new `enum`, `Direction`. This is where we specify the north/east/south/west direction that we would like to anchor to.
- Implement new protocol, `Anchor`. This extends `Annotation` to include a direction and a float buffer (space desired between the annotation and the anchor point). We also require a `resolve` function that will deal with the anchor directions and update the parameters of the annotation accordingly.
- Have our existing `Box` and `Text` annotations also conform to `Anchor`. Now, we could just replace this with just conforming to `Anchor` as it extends `Annotation`, however I think that this definition is more verbose.
- Update the `Arrow` annotation to support anchoring of any annotation that supports it. Also included an option to override the anchor and force to draw at passed coordinates if the user desires.
- Update the arrow tests to support this feature- no more "guess and check" to find the locations to pass anymore!

So what does this mean for the future of annotations in SwiftPlot?

- Adding annotations will be pretty straightforward. If we don't want it to support anchoring, that's fine. Otherwise, conform it to `Anchor` and we're good to go.
- This introduces the possibility to anchor the annotations to anything we want, not just the ends of an arrow. We can pass any point we'd like to the `resolve` function and it will relocate itself. We can make it anchored to the origin, ends of lines in line graphs, etc... the list is endless.

Anyways, hope you guys like it. I'm definitely open to any suggestions for improvements or new ideas. Feel free to check it out.

William


WilliamHYZhang:
Oooof. PR build is failing... I'll look into it.

WilliamHYZhang:
Fixed, @karwa when this is merged may you rerun quartz tests when you have time? No hurry.

WilliamHYZhang:
@KarthikRIyer resolved merge conflicts

KarthikRIyer:
Great work @WilliamHYZhang. I really like how you've approached the problem! I have some suggestions:
- Instead of `Anchor`, I think `AnchorableAnnotation` makes more sense.
- Instead of buffer, does margin seem more suitable, because when I first read Float buffer, I thought about a Float array and was wondering how that was required.
- In the example It would be nice if the text is anchored at the middle of it's height.

KarthikRIyer (commented):
Great work @WilliamHYZhang. I really like how you've approached the problem! I have some suggestions:

- Instead of Anchor, I think AnchorableAnnotation makes more sense.
- Instead of buffer, does margin seem more suitable, because when I first read Float buffer, I thought about a Float array and was wondering how that was required.
- In the example It would be nice if the text is anchored at the middle of it's height.

WilliamHYZhang:
@KarthikRIyer should be good to go.

KarthikRIyer:
Looks good @WilliamHYZhang. Good job!

PR #102, opened 2020-01-19
Add Arrow Features

WilliamHYZhang:
Introducing some new features to our beloved `Arrow` annotation.

Some code reworkings to:
- Make a new `HeadStyle` enum for the existing head styles, making it easier to add more in the future.
- Make two new helper functions to draw the arrow body and the arrow head, allows for more flexibility and generalization.

New features/styles include:
- Option to specify points where the arrow body line should go through, allows for curved/angled arrow bodies with the addition of `linePoints` array.
- Option to have a double-headed arrow with the addition of `isDoubleHeaded`.

There will probably be some merge conflicts after #100 is merged... I'll resolve those when that happens.


KarthikRIyer:
Good start @WilliamHYZhang. It would definitely be nice to have more arrow styles. But this seems like a good framework making it easier to add more styles.
Although I'm not sure your way of drawing the body is the right way to go forward. This is because if we want a curved arrow we'd like the annotation to handle everything. Here the user will have to compute all the curve points and pass it in. At the moment I'd suggest that the body be rolled back to the older version. We need to think a bit more about it's implementation.

KarthikRIyer (commented):
Good start @WilliamHYZhang. It would definitely be nice to have more arrow styles. But this seems like a good framework making it easier to add more styles.
Although I'm not sure your way of drawing the body is the right way to go forward. This is because if we want a curved arrow we'd like the annotation to handle everything. Here the user will have to compute all the curve points and pass it in. At the moment I'd suggest that the body be rolled back to the older version. We need to think a bit more about it's implementation.

KarthikRIyer:
Also could you add a new test for this?

WilliamHYZhang:
The intuition behind the addition of `linePoints` is not really for the curve (sorry I was a bit misleading), more so that if the user wants angled arrows such as in the shape of an L, etc. I'll roll it back for now.

Regarding the arrow styles, I tried to add all specified in the matplotlib documentation [here](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.patches.ArrowStyle.html), except for those such as Bracket, Wedge, which are not really arrows and would be added as a new annotation structure. I'll add some more head styles today.

WilliamHYZhang:
@KarthikRIyer should be good to go.

KarthikRIyer:
I like this. Just one more change. In the examples could you make the arrow stroke thickness a little lesser? It seems very thick now.

WilliamHYZhang:
Sure, I'll do that and resolve the conflicts.

WilliamHYZhang:
@KarthikRIyer Just checked, the arrow thickness is now set to the default width of 5 for all tests (this is what it was like before with the exception of the wedge arrow, which I changed to default and made a little less long). Resolved merge conflicts and also updated tests to use the new anchor functionality, let me know if anything else needs to be changed. Thanks!

KarthikRIyer:
Looks good. Thanks @WilliamHYZhang !

PR #106, opened 2020-01-23
Polar chart (in progress)

areebg9:
.addSeries and .addFunction work, only issue is when I tried to implement the `draw_empty_circle` function in the renderers.


KarthikRIyer:
@Qwerty71 the reason you're getting this error is that you haven't defined the draw_empty_circle(float, float, float, void const*) method in CPPAGGRenderer.cpp file. The function you've defined doesn't take in the void const* parameter. Take a look at line 535 in that file for example. I'll try to explain a bit how this works.

AGGRenderer is written in CPP. Although the swift package manager can compile CPP code, we can't call CPP functions directly from Swift(But the support for that is coming. I have seen a prototype at a demo in one of the open design meeting of S4TF ). We can call C functions directly from swift. So what we do here is, wrap the CPP functions in C functions using a bridging header. 

Look at line 524 of the same file: The initializePlot function initializes an object of AGGRenderer and returns a pointer to the object in memory. Everytime we need to call a function of that object we pass it in as a void pointer, cast it to a Plot pointer and then call the function. This is how it's done in all the drawing functions like draw_solid_rect or draw_solid_circle.

Hope this helps.

KarthikRIyer:
I don't remember. Are you on Linux or mac?  If you're on mac could you implement the complete function to draw the circle in QuartzRenderer? If you're on Linux could you atleast add an empty function so that it conforms to the Renderer protocol?

areebg9:
Hey, I've got one last question - how do I create a strokeWidth or border of a circle? I see the `stroke` variable being used for drawing lines, but I'm not sure whether I can use it to outline a circle.

KarthikRIyer:
You should be able to do the same with a circle. You need to convert the path to a stroke. See how it is done in draw rect function.

areebg9:
Hey,

How do we get the size of the graph / layout? I'm trying to get input a radius to the function `drawEmptyCircle`, but I can't figure out how to do it (for normal points, I use `convertCoordinates` so it scaled properly with the graph).

areebg9:
I've gotten this last issue to work, now I just need to draw a circle, transparent in the fillColor, but with a color stroke.

WilliamHYZhang:
@Qwerty71 please either update the protocol or the conforming struct to resolve build errors.

areebg9:
It is still in progress (as per the title of this PR) :)

WilliamHYZhang:
@Qwerty71 so far it's looking good though, keep it up :+1:

areebg9:
Thanks! I have everything working now (thanks @KarthikRIyer for the help), I'm just automatically making the circles for the gridlines.

areebg9:
Hey, could anyone help me with this issue? How do I clip the graph so that any extraneous lines/circles do not appear? An example of this issue can be found [here](https://colab.research.google.com/drive/1ZBPo2yRulZcrSDdQx1wizCRG9VkoRt6f). Help would be much appreciated!

WilliamHYZhang:
@Qwerty71 you could include a check to only graph the circles that fall INSIDE the graph perhaps.

WilliamHYZhang:
Clipping the circles will take significantly more work: you'll have to split them up and graph individual arcs, or maybe include an overlay outside the graph which isn't ideal.

KarthikRIyer:
@Qwerty71 could you check the allow edits from maintainer option in your PR. I'll fix the build fail by adding the code for QuartzRenderer.
Also I agree with @WilliamHYZhang. Instead of clipping the circles just calculate the maximum radius of the circle that can come in your figure.

areebg9:
Yes. The only problem is that the figure is rectangular, rather than a
square, which means that his implementation would fail to cover the
external radii. I had the idea of fixing the graph to become a square (have
the axes be the same for both), and then I can get it to work. If you have
any suggestions in implementing that, please let me know :)

KarthikRIyer:
Actually you'd have to re-implement the plot instead of reusing Line Graph code. Try taking a look at the line graph code to understand the flow. Let me know if you have any doubts. Then we can move on to implementing this plot from scratch.

Line graph takes in width and height. You can instead have just one parameter for Polar Chart.

KarthikRIyer:
Also we do not want the rectangle border. The border should be circular. Take a look at how matplotlib polar plot results look like. https://matplotlib.org/gallery/misc/transoffset.html#sphx-glr-gallery-misc-transoffset-py

areebg9:
Got it. I can implement the arc function, but if we want to make the borders circular then there is no need. In fact, if we make the borders circular, then there is no need for a background layout at all - all I need to do is draw the circles. Would you like me to include the degree measure in the plot?

KarthikRIyer:
Yes, it'd be nice to include the degree measure. Try to replicate the example I sent above, as much as possible.

areebg9:
Sorry, I've been rather busy lately. I will finish this task and implement Polar Plot when I can. So far, I need to 
a) make the grids circles (and the border, as per matplotlib style)
b) add degree measures
c) (if necessary - it would be convenient but is no longer needed) create function to generate arcs.

KarthikRIyer:
No worries @Qwerty71. This isn't an urgent task. Take your time. :D

WilliamHYZhang:
Nice to see progress on this! Keep it up 👍

PR #107, opened 2020-01-23
Function to draw solid ellipses

boronhub:
Added function to draw solid ellipses in AGGRenderer and SVGRenderer.

### DIFF
--- a/Sources/SwiftPlot/Renderer.swift
+++ b/Sources/SwiftPlot/Renderer.swift
@@ -129,6 +129,23 @@ public protocol Renderer: AnyObject{
                          radius r: Float,
                          fillColor: Color)
 
+
+    /*drawSolidEllipse()
+    *params: center c: Point,
+    *        radius rx: Float,
+    *        radius rx: Float,
+    *        fillColor: Color,
+    *        isOriginShifted: Bool
+    *description: Draws an ellipse with specified fill color, center and radii
+    *             This function can operate in both coordinate systems with and
+    *             without shifted origin.
+    *             This is decided by the boolean parameter isOriginShifted.
+    */
+    func drawSolidEllipse(center c: Point,
+                          radius rx: Float,
+                          radius ry: Float,
+                          fillColor: Color)
+
     /*drawSolidTriangle()
     *params: point1: Point,
     *        point2: Point,
--- a/Sources/AGGRenderer/AGGRenderer/AGGRenderer.swift
+++ b/Sources/AGGRenderer/AGGRenderer/AGGRenderer.swift
@@ -123,6 +123,22 @@ public class AGGRenderer: Renderer{
                         agg_object)
     }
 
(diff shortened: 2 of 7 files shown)


boronhub:
@KarthikRIyer any idea how to fix the QuartzRenderer error ?

WilliamHYZhang:
@boronhub comment out the Quartz Renderer tests if you're not using an Apple device.

Sources/SwiftPlot/Renderer.swift
```diff
@@ -129,6 +129,23 @@ public protocol Renderer: AnyObject{
...
+    /*drawSolidEllipse()
+    *params: center c: Point,
+    *        radius rx: Float,
+    *        radius rx: Float,
+    *        fillColor: Color,
+    *        isOriginShifted: Bool
+    *description: Draws an ellipse with specified fill color, center and radii
+    *             This function can operate in both coordinate systems with and
+    *             without shifted origin.
+    *             This is decided by the boolean parameter isOriginShifted.
```
WilliamHYZhang, line 142: Could you add support for `isOriginShifted` as provided in the description?

WilliamHYZhang (commented):
Great stuff @boronhub

KarthikRIyer:
You'll need to implement the function in QuartzRenderer too. This is because QuartzRenderer needs to conform to the Renderer protocol.
You'll be able to do this only if you are on a mac. If not, could you allow edits from maintainer option in your PR. Then I can add the relevant code to fix the build.

boronhub:
Done!

KarthikRIyer:
I don't have ready access to a mac (I did have one an hour ago). So I'll add the code as soon as I get my hands on a mac.

Sources/SwiftPlot/Renderer.swift
```diff
@@ -129,6 +129,23 @@ public protocol Renderer: AnyObject{
...
+    */
+    func drawSolidEllipse(center c: Point,
+                          radius rx: Float,
+                          radius ry: Float,
```
odmir, line 146: Could you change the name of the external labels for the radius? For someone using this function it will not be clear which parameter is the x or y radius unless they go look at this implementation.

odmir:
I think this should work:
```swift
public func drawSolidEllipse(center c: Point,
                             radius rx: Float,
                             radius ry: Float,
                             fillColor: Color) {
    let ellipse = CGMutablePath()
    ellipse.addEllipse(in: CGRect(x: CGFloat(c.x-rx), y: CGFloat(c.y-ry), width: CGFloat(rx*2), height: CGFloat(ry*2)),
                       transform: CGAffineTransform(translationX: CGFloat(xOffset), y: CGFloat(yOffset)))
    context.setFillColor(fillColor.cgColor)
    context.addPath(ellipse)
    context.fillPath()
}
```

boronhub:
Commits + fixes pushed by @KarthikRIyer through #111

PR #110, opened 2020-01-26
Turn on quartz tests for annotations

odmir:
- Used the new nifty `renderAndVerify` helper function in almost all tests.
- Consequently turned on Quartz testing for `Annotation`s.
- Updated `verifyImage` to check if two images are equal iff both images actually exist. This is to reduce visual noise in the tests' output.


KarthikRIyer:
Thanks @odmir! Merging.

PR #113, opened 2020-01-27
Performance improvement to `recalculateBins`

odmir:
Around the end of last year I noticed that the time the tests were taking was starting to become a burden, especially when doing quick iterative work. After investigating I saw that a big chunk of it was due to the increase in histogram tests at that time, which were really exercising one specific method which took a big slice of the testing time and is called several times for each test, at least twice per output file. Due to this, I decided to change the algorithm used in `recalculateBins` to speed it up considerably, and consequently the testing time as well.

The algorithms are documented in the code. I tried giving them big O notation in the comments for performance comparison, but I have never done this before, so if it is stupidly wrong please tell me 😜.

Regarding the performance wins when executing all of the tests on my machine, the new tests report this:

before:
Screenshot 2020-01-27 at 04 27 16
after:
Screenshot 2020-01-27 at 04 28 25
they are two tests testing two different algorithms, one used when the data is sorted and another for when its not;

and when executing all the tests:

|                | Without New Tests | With New Tests |
|----------------|---------------|------------|
| master         |    ~ 10.1 s   |  ~ 13.7 s  |
| new algorithms |    ~ 7.3 s    |   ~ 8.1 s  |

The new tests added by this PR add some time to the overall testing time, but it is still an improvement on how it was before, even without the new tests.

Now the time `recalculateBins` takes is basically negligible.

At the time the improvements in testing speed seemed better because there are more tests now and it looks less impressive 😜.

Please tell me about any issues you encounter. I only wanted to test that specific method, but it was marked `private` so I created another internal method that just calls it for testing purposes. I don't know if there is a better approach other than measuring the performance of that private method indirectly by measuring the whole rendering process, or just by changing the access of that method back to `public` like it was before. @KarthikRIyer @karwa.


Sources/SwiftPlot/HistogramSeries.swift
```diff
@@ -1,5 +1,7 @@
-public struct HistogramSeries<T> where T: Comparable {
-    public var data = [T]()
+public class HistogramSeries<T> where T: Comparable {
```
karwa: Why has this become a class?

karwa (commented):
I need to look at this in more detail. I will note for complexity purposes that we expect n >> m. Good values can be around sqrt(n).

Sources/SwiftPlot/HistogramSeries.swift
odmir: Oh I’ll revert it back to a struct, I completely missed that after merging from master, I will also take a look at the rest to see if anything else got changed that shouldn’t have.

KarthikRIyer:
I like this approach! I hadn't paid much attention to the time complexity of this function although I knew it wasn't good. I tried to improve it once but messed it up and forgot about it later 😅. 
This looks good to me. But I'm not sure of the O() notation for the sorted case. I think it would be O(n) generally (for the case Karl specified above).

odmir:
Ok I've changed the `class` back into a `struct`.
Regarding the O notation in the sorted case it should become O(n) and in the unsorted case it should be O(n log(m)) as well assuming n >> m right?
If we assume that in general m ~= sqrt(n) then it should be something like: O(n) if sorted; O(n log(√n)) = O(n log(n)/2 ) if unsorted; right?

Sources/SwiftPlot/Histogram.swift
```diff
@@ -310,27 +310,82 @@ private extension Histogram {
...
+            /// Performance: O(n + m). n: `data.count`, m: `binFrequency.count`.
+            ///   This algorithm iterates through `data` and `binFrequency` only once.
+            ///   It sets `binFrequency` and checks if it is the maximum frequency `binFrequency.count` amount of times.
+            var dataTail: Int = 0
+            var dataHead: Int = 0
+            let dataEndIndex = series.data.endIndex
+            var binIndex: Int = 0
+            let binEndIndex = binFrequency.endIndex
+            var xUpperLimit = binStart + T(binIndex + 1) * binInterval
+            while true {
```
karwa: I'd prefer if this was written as `while binIndex < binEndIndex`. It makes it easier to read and to check that the code doesn't loop forever.

Sources/SwiftPlot/HistogramSeries.swift
```diff
@@ -1,5 +1,7 @@
 public struct HistogramSeries<T> where T: Comparable {
-    public var data = [T]()
+    public var data = [T]() { didSet { isSorted = false } }
```
karwa: Rather than doing this, we could make data `public private(set)` and add a setter which asks if the data is sorted. That way users don't lose this performance win if replacing the data with other, already-sorted data.

Sources/SwiftPlot/HistogramSeries.swift
```diff
@@ -11,6 +13,7 @@ public struct HistogramSeries<T> where T: Comparable {
                 color: Color,
                 histogramType: HistogramSeriesOptions.HistogramType) {
         self.data = data.sorted()
+        self.isSorted = true
```
karwa: It would be better to make this a parameter in the initialiser, rather than just assuming that all data is already sorted. We never document anywhere that this data should be sorted.

karwa (changes requested):
A couple of minor comments, but overall +1! 👍 

Unfortunately, XCTest performance tests are kind of useless for CI, because we can't set the baseline measurements. I would really like them, but there's not much point to adding them at this time :(

Sources/SwiftPlot/HistogramSeries.swift
KarthikRIyer: But while setting the data we sort it. 

Self.data = data.sorted()

karwa: We happen to do that today, but it's fragile to be relying on histogram always pre-sorting the data - maybe somebody will add a method to histogram and forget to sort the data first. There would be no error, just worse performance.

Also, this initialiser is public and `Histogram.histogramSeries` is publicly settable; there is nothing stopping anybody replacing the instance entirely with `myHistogram.histogramSeries = HistogramSeries(data: [4, 3, 2, 1], ...)`. If we assume that it's always sorted, we would display an invalid graph.

odmir: > Also, this initialiser is public and Histogram.histogramSeries is publicly settable; there is nothing stopping anybody replacing the instance entirely with myHistogram.histogramSeries = HistogramSeries(data: [4, 3, 2, 1], ...). If we assume that it's always sorted, we would display an invalid graph.

Ok I think you might be misunderstanding where `self.isSorted = true` is defined. The example you gave would lead to correct behaviour, and we would not display an invalid graph. This is because we sort inside the `HistogramSeries` initialiser at the same time that we set `isSorted = true`, not in `Histogram`'s initialiser or anywhere else outside `HistogramSeries`'s initialiser. So `HistogramSeries(data: [4, 3, 2, 1], ...)` sorts the data before storing it and then sets `isSorted = true`. Anyone that touches `data` directly afterwards would invalidate this flag, thus setting `isSorted = false` implicitly.
I do agree on having a separate setter with a parameter to ask if the data is sorted though.

I guess the question is if we want to keep the current behaviour of master, which is to sort any data that comes through the initialiser of `HistogramSeries` or if we want to stop doing that every time and instead expose the `isSorted` parameter through that same initialiser, giving the option of pre-sorting to `Histogram` and others while giving the ability, at instance creation, to inform `HistogramSeries` that the data passed in is already sorted.

Hopefully this makes sense and I did not misunderstand anyone!

karwa: > This is because we sort inside the HistogramSeries initialiser

Ah, my mistake - it's actually on the very line above this one 🤦‍♂

I guess in the long run it makes to give the user an option to not sort the data (since apparently that's something we support), but for now it's fine not to do the initialiser parameter.

I'm still not entirely happy with assuming that any modification disrupts the data's sorting. This isn't an entirely easy thing to solve though. If we replaced direct setting with a method, things like `.append` would become unavailable. The only way to keep those methods is to use a closure:

```swift
mutating func modifyData(preservesSorting: Bool, _ closure: (inout [T])->Void) {
  closure(&data)
  isSorted = !preservesSorting
}
```

But that's not entirely obvious, either - because we automatically sorted the data for the user, they might not be aware that by "preserves" sorting, we mean that they won't un-sort previously sorted data. Which is to say: if they gave us unsorted data, then make some random edits, they might say that they "preserve" sorting because the data wasn't sorted before and still isn't sorted afterwards.

Alternatively, we could not try to be clever, and make `isSorted` public instead. Turn it off on changes by default, but give the user the ability to get that performance back if they make the effort to sort their data.

Sources/SwiftPlot/HistogramSeries.swift
odmir: If we want to protect a little against ourselves would it be a good idea to do what you suggest and also leaving the didSet?
My thinking is that, if we ever were to add more mutating methods to `HistogramSeries` in the future, and by chance forget to use the correct setter (lets call it `setData(_: [T], sorted: Bool)`, and instead we set `data` directly without setting `isSorted` to a correct value as well, it could lead to very subtle bugs where `isSorted` should be false after such a method but because it stayed true now the algorithm throws garbage.
If we would keep the `didSet`, even with a private(set), at least in these cases it wouldn't lead to mysterious bugs in some rare cases, but instead lead to a worse performance. I can see though why the better outcome in this case could be to generate a garbage graph, instead of a good one that was generated slower, in terms of how quickly the underlying bug would be caught.

Sources/SwiftPlot/HistogramSeries.swift
karwa: Actually - there is probably an even better option: don't sort the data.

There are two main reasons for this:

- **Usability**: As a user, I just created a histogram with some data. Now I want to access or modify some of that data, but it has unexpectedly changed from what I gave it. Everything is in a different location, so I can't reliably process it.

- **Performance**: Sorting is an `O(n log(n))` operation, which is more than we save by having the data sorted. Also, it can lead to additional copies of the data (another `O(n)` operation). And then if a user modifies the data after creation, we throw away all that work we did by sorting anyway. We made it harder for them to use, and it wasn't even for a good reason.

Would you mind benchmarking it and seeing what happens if we just never sort the data?

odmir: I can see how it can quickly become hairy... My initial implementation was based on the fact that the data was already being sorted every time in the initialiser, so I though "why not take advantage of that fact when possible?".My aim was not to introduce a new `isSorted` property that everyone else now would have to be careful with so I decided to not include any public api that exposed this, and only added the necessary code to make sure that the algorithm doesn't break (the `didSet`). In a way `isSorted` doesn't really get through the idea I wanted, it should be more like `is(Definitely|Undeniably)Sorted` since I only cared that the fast algorithm never runs when `data` actually isn't sorted, with no big harm if `data` is sorted but the algorithm that runs is the slower one.

I don't know if you want to continue discussing solutions to possible public facing apis here, that don't run into the disadvantages you mention above.

What do you think of doing the following for right now:
- set `isSorted = false` on any modification to `data`
- no `private(set)` so we can still use `.append` and such directly
- we could still have a separate public setter (`setData(_:sorted:)`) to give performance minded users the ability to say if their new data is sorted or not (this could also come later)
- keep sorting the data and setting `isSorted = true` in the initialiser without exposing an `isSorted` parameter on it

Then in another PR/Issue we could discuss how we could expose the standard apis of `data` while letting the user specify sortedness without the drawbacks you mentioned. This includes discussing the third point above (`setData`) if we decide to not include it right now in this PR.

karwa: I think first we should consider if sorting the data is even worthwhile. If our algorithm didn't even need it in the first place, that's a clear an obvious removal of unnecessary work, with no API awkwardness to worry about.

We can then consider if it makes sense, if the user already _happens_ to have sorted data, to allow them to mention that and benefit from a faster algorithm.

odmir: oh sorry, I took my time writing and only now I see another comment of yours came through before mine

odmir: > Actually - there is probably an even better option: don't sort the data.
> 
> There are two main reasons for this:
> 
> * **Usability**: As a user, I just created a histogram with some data. Now I want to access or modify some of that data, but it has unexpectedly changed from what I gave it. Everything is in a different location, so I can't reliably process it.
> * **Performance**: Sorting is an `O(n log(n))` operation, which is more than we save by having the data sorted. Also, it can lead to additional copies of the data (another `O(n)` operation). And then if a user modifies the data after creation, we throw away all that work we did by sorting anyway. We made it harder for them to use, and it wasn't even for a good reason.
> 
> Would you mind benchmarking it and seeing what happens if we just never sort the data?

ok yes that makes sense, I will benchmark it

odmir: > I think first we should consider if sorting the data is even worthwhile. If our algorithm didn't even need it in the first place, that's a clear an obvious removal of unnecessary work, with no API awkwardness to worry about.
> 
> We can then consider if it makes sense, if the user already _happens_ to have sorted data, to allow them to mention that and benefit from a faster algorithm.

That also makes sense 👍 

Edit:
Actually just from the usability perspective it seems obvious we should really not be sorting in the first place.

I can see two logical avenues:
1. We make `data` internal, requiring users to make sure the data is all processed before creating a `HistogramSeries` and requiring them to create a new one for data that was further processed;
2. We leave it public so users can process the data in `HistogramSeries` whenever they want.

If we go with 1 we are free to do whatever we want with `data` and don't have to worry with exposing anything related to it, allowing us more flexibility.
If we go with 2 then our hands are basically tied and we can't touch `data` (be it for whatever optimisation we might want to do in the future or other reasons) in any way because as a user I would think it is unacceptable for my data to be mangled after being stored in `HistogramSeries` before trying to process it in some way.

In my opinion I think 1 actually makes more sense... not because it would be the only option that would allow us sorting the data whenever we want, but because I don't think of a `HistogramSeries` as the appropriate thing to use to do processing of data. I would rather process the array of data I already have and create a new `HistogramSeries` whenever I want to make a `Histogram`.
This also gives us flexibility, in theory we could exchange the underlying data array with some other data structure for some reason and also we don't have to make promises that we won't change the user's data without them knowing.

karwa: Option 1 would differ from the other plots, and ultimately it all depends on whether there _are_ any such optimisations (histograms are a well-explored domain, and I couldn't find anything), what the cost is to convert the data to that format, and what the gains are. Given that any copy will cost us _at least_ `O(n)`, I think it's unlikely that any data structure exists that would give us back the cost.

> not because it would be the only option that would allow us sorting the data whenever we want

Like I said, it's not clear that sorting the data actually _is_ a win, whenever we do it. It costs up to `O(n log n)`, all so we can turn another `O(n log n)` operation to only `O(n)`. We should see an overall performance boost by cutting all of that out and only doing a single `O(n log n)` operation on the unsorted data. In other words: it should be _even faster_ (or at least more scalable) than what you have right now.

If the user already has sorted data from somewhere - fine, we can use that information. That's (essentially) zero-cost for us, and you've shown that you can extract gains from that knowledge.

If we encounter some compelling reason to use another data structure, we'll consider what to do at that time. As it is, I don't see the harm.

odmir: Yes, I completely agree with you, what I meant with that was that my opinion to choose the first option wasn't based on being able to sort or not ahah. And with that I have the results of the experiment!

First to make sure I am not making some mistake please validate what I did:
Create new temp initialiser that does not sort and sets `isSorted = false`:

```swift
public init(dataNotSorted: [T],
            bins: Int,
            label: String,
            color: Color,
            histogramType: HistogramSeriesOptions.HistogramType) {
    self.data = dataNotSorted
    self.isSorted = false
    self.bins = bins
    self.label = label
    self.color = color
    histogramSeriesOptions.histogramType = histogramType
}
```

Then the two tests:

```swift
// time of fast algorithm + necessary sort
func testPerformanceHistogramCreateSeriesAndRecalculateBins_Sorting() throws {
    let histogram = Histogram<Float>(isNormalized: false, enableGrid: false)
    
    measure {
        let histogramSeries = HistogramSeries(data: histogram_step_values, bins: 50, label: "HISTOGRAM PERFORMANCE `addSeries` & `recalculateBins`", color: .black, histogramType: .bar)
        histogram.testRecalculateBins(series: histogramSeries, binStart: 40, binEnd: 160, binInterval: (160-40)/Float(histogram.histogramSeries.bins))
    }
}
// time of slow algorithm
func testPerformanceHistogramCreateSeriesAndRecalculateBins_NonSorting() throws {
    let histogram = Histogram<Float>(isNormalized: false, enableGrid: false)
    
    measure {
        let histogramSeries = HistogramSeries(dataNotSorted: histogram_step_values, bins: 50, label: "HISTOGRAM PERFORMANCE `addSeries` & `recalculateBins`", color: .black, histogramType: .bar)
        histogram.testRecalculateBins(series: histogramSeries, binStart: 40, binEnd: 160, binInterval: (160-40)/Float(histogram.histogramSeries.bins))
    }
}
```

assuming I'm not overlooking anything the results are:
Screenshot 2020-01-29 at 20 10 16

which is a pretty clear win for not sorting in the first place! So it is now time to simplify 😂.

Nonetheless I still think option 1 would make sense, not exposing the underlying array at all (which in my opinion is an implementation detail), unless there is a compelling enough reason, and reconsider the same for all the other plot types, but I haven't actually looked too hard at the rest nor have I though enough about what that would mean overall.

odmir: I don't think the difference between the fast and slow algorithms, when a user gives an already sorted array, justifies the added complexity of two algorithms and maintaining an `isSorted` property, when the need comes we can think about giving the user the option of giving sorted data.

KarthikRIyer: The above discussion has been illuminating for me! I never thought about the cost of sorting the data prior to binning.

What does **(1% worse)** actually mean? What is that time being compared to?

One more thing unrelated to this dicussion: matplotlib return the bins and the histogram values. Should we consider adding a function to do something like this?

odmir: That time is being compared to a baseline set previously, in my case by measuring. The 1% worse just means that this time it was 1% slower than the baseline.

Regarding exposing the bins and the histogram values, in my opinion we should come up with use cases for that and weight if it makes sense/is desirable first.
Screenshot 2020-01-29 at 20 50 24

karwa: We can consider hiding the underlying array if/when a need arises. Maybe it will be for this plot, maybe for some other plot. I'm open to it if there's a compelling use-case.

In the mean time, those tests look good.

odmir:
> Unfortunately, XCTest performance tests are kind of useless for CI, because we can't set the baseline measurements. I would really like them, but there's not much point to adding them at this time :(

@karwa is there some solution that you know of? Also is there any way to disable some tests, or separate them into another "module" so that CI only runs the non-performance related tests for now?

odmir:
Ok the build seems to be failing because some other behaviour is dependent on the data being sorted, it would have broken sooner or later, at least as soon as someone tried to plot a histogram after manually changing data.

odmir:
I had to put a workaround in `layoutData` because code over there was getting the max and min X by getting the first and last elements from `data` causing trouble later on. I added a comment documenting the workaround and the why as well as a `FIXME` because we are not handling when the user uses an empty array, and a "performance" `TODO` because now we are going through the entire `data`, however many series the histogram has.

karwa:
> is there some solution that you know of? Also is there any way to disable some tests, or separate them into another "module" so that CI only runs the non-performance related tests for now?

I don't know of a solution. The problem appears to be that Xcode records and saves baseline measurements as plist files, and that the package manager has no knowledge of those files or any parameters where you can specify them.

In any case, we would likely _want_ those tests to be run on CI. It's less important that CI is fast, and more important that it is accurate and doesn't let us accidentally commit a major performance regression without realising it.

> I had to put a workaround in layoutData because code over there was getting the max and min X by getting the first and last elements from data causing trouble later on. I added a comment documenting the workaround and the why as well as a FIXME because we are not handling when the user uses an empty array, and a "performance" TODO because now we are going through the entire data, however many series the histogram has.

Well that's unfortunate. What does this do to performance? Is there still an overall improvement by disabling sorting?

Just by thinking about it, I would assume there _is_ still an improvement, at least in our tests. Our data is defined as static `let` constants, so when we sort it we'll be making a copy in `O(n)` before doing the sorting in `O(n log(n))`. By getting the min/max without mutating, we can turn that in to a single `O(n)` operation.

Sources/SwiftPlot/HistogramSeries.swift
KarthikRIyer: Regarding exposing the bins and histogram weights, I talked to someone who works in ML. These values come in useful during classification tasks where you need to categorize continuous values into discrete values.

karwa: @KarthikRIyer that's more-or-less the same data that's in Histogram's `DrawingData` type, isn't it? We could definitely support querying that directly.

We could also expose that through something similar to MATLAB's [`histcounts` API](https://de.mathworks.com/help/matlab/ref/histcounts.html).

This gets more interesting if we introduce algorithms to automatically determine a good number of bins. MATLAB provides [lots of options](https://de.mathworks.com/help/matlab/ref/histcounts.html#buijikm-1-BinMethod), and there's lots of research with alternative algorithms.

KarthikRIyer: > @KarthikRIyer that's more-or-less the same data that's in Histogram's DrawingData type, isn't it? We could definitely support querying that directly.

@karwa yes.

Yeah, this idea for auto number of bins is very interesting! We can definitely get that implemented. I'd love to work on this, but I'm not sure how soon I can get to it.

odmir:
Yes it is still better. For the benchmarks I created:
1. one branch from a commit where we had: sorting, faster algorithm and fast maxX, minX;
2. another branch from latest commit with: no sorting, fast algorithm and slower maxX, minX.

From both branches I removed the performance tests because they only exercised the algorithm and sorting parts but never touched the maxX and minX.

Then I measured only the `HistogramTests` test suite:

Branch|Average Time (s)
:-:|:-:
1|1.648
2|1.567

so we are better now with an improvement of about 80 ms.

odmir:
and compared to master:

Branch|Average Time (s)
:-:|:-:
2|1.567
master|4.323

an improvement of 2.756 s.

odmir:
If everyone agrees this is ready to be merged!

@KarthikRIyer @karwa

karwa:
Could you maybe move the test to a new test suite? Maybe `PerformanceTests`? Then we could add performance tests for other plots there, and you can easily exclude them all if you don't want to run them.

odmir:
There you go! 😊

karwa:
Nice. Looks good to me 👍

KarthikRIyer:
Great! Thanks @odmir . Merging.

PR #114, opened 2020-01-28
Create SwiftPlot_Ubuntu.yml

KarthikRIyer:
I've added a new Github Action that builds SwiftPlot on Ubuntu. Please take a look and let me know of any required changes.

@karwa @odmir

### DIFF
--- a/.github/workflows/SwiftPlot_Ubuntu.yml
+++ b/.github/workflows/SwiftPlot_Ubuntu.yml
@@ -1,0 +1,28 @@
+name: SwiftPlot Ubuntu
+
+on:
+  push:
+    branches: master
+  pull_request:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install Freetype
+      run: sudo apt-get install libfreetype6-dev
+    - name: Install Swift Dependencies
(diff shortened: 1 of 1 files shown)


karwa:
Looks good! It's really nice to have CI validate both platforms. I'm also keeping an eye on Windows support; at least [one project](https://github.com/SDGGiesbrecht/Workspace/blob/master/.github/workflows/Workspace%20Validation.yaml#L34) is doing it, but it's not straightforward.

One thing that would be good is to upload the generated files, like we do on the Mac. GitHub doesn't have a [usage limit](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#usage-limits) for it (limits are all about concurrent jobs), so why not?

KarthikRIyer:
@karwa done.

Is there support for SPM in Windows?

karwa:
@KarthikRIyer I honestly don't know.

KarthikRIyer:
There's this: https://github.com/SwiftForWindows/SwiftForWindows
But it doesn't provide SPM. To build we have to use a build.json file instead of Package.swift. So I think it will be difficult to get it to work with windows.

karwa:
I think that project is quite old; that's not the latest state of Swift on Windows.

[This document](https://github.com/apple/swift/blob/master/docs/WindowsBuild.md) in the official repo seems to suggest that SPM is supported.

PR #121, opened 2020-03-08
Allows SVGRenderer to export the result from memory

### LINKED ISSUE #120: Allow SVGRenderer to export the result from memory
Instead of saving, it would be cool to have somehting like:
```swift
let svg = svg_renderer.svg() // Result as plain string
```
Implementing would be very easy:
```swift
extension SVGRenderer {
    /// Returns the content of the SVG generated by the renderer
    func svg() -> String {
            // Build the document.
            let header = #"<svg height="\#(imageSize.height)" width="\#(imageSize.width)" version="4.0" xmlns="http://www.w3.org/2000/svg" xmlns:xlink= "http://www.w3.org/1999/xlink">"#
                + "\n" + #"<rect width="100%" height="100%" fill="white"/>"#
            let font = #"<defs><style>@import url("https://fonts.googleapis.com/css?family=\#(fontFamily)");</style></defs>"#
            let image = header + "\n" + font + "\n" + lines.joined(separator: "\n") + "\n</svg>"
            return image
    }
}
```

arguiot:
Fixes #120


KarthikRIyer:
Thanks a lot for your contribution @arguiot. Appreciated!

